PR #6, opened 2017-11-28
Set up seed data & deploying to Heroku

irisfaraway:
https://eaflood.atlassian.net/browse/WC-121

The app relies on both registrations and users being created by the existing service. We'll need some seed data for existing users and their registrations so we can test the app independently.

This PR will also cover any config changes required to deploy to Heroku for testing.

### DIFF
--- a/app/models/conviction_search_result.rb
+++ b/app/models/conviction_search_result.rb
@@ -4,13 +4,12 @@ class ConvictionSearchResult
   embedded_in :registration
   embedded_in :keyPerson
 
-  # TODO: Confirm types
-  field :matchResult,     type: String
-  field :matchingSystem,  type: String
-  field :reference,       type: String
-  field :matchedName,     type: String
-  field :searchedAt,      type: DateTime
-  field :confirmed,       type: Boolean
-  field :confirmedAt,     type: DateTime
-  field :confirmedBy,     type: String
+  field :matchResult, as: :match_result,        type: String
+  field :matchingSystem, as: :matching_system,  type: String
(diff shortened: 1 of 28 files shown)


irisfaraway:
Now deploying to https://waste-carriers-renewals.herokuapp.com/

app/models/conviction_search_result.rb
```diff
@@ -5,12 +5,12 @@ class ConvictionSearchResult
...
   # TODO: Confirm types
-  field :match_result, type: String
-  field :matching_system, type: String
-  field :reference, type: String
-  field :matched_name, type: String
-  field :searched_at, type: DateTime
-  field :confirmed, type: Boolean
-  field :confirmed_at, type: DateTime
-  field :confirmed_by, type: String
+  field :matchResult,     type: String
```
Cruikshanks: There are a number of changes to set field names to match what the current document looks like in the db. However we can keep our more ruby like syntax using field aliases

```ruby
field :matchingSystem, as: :matching_system, type: String
```

<https://docs.mongodb.com/mongoid/master/tutorials/mongoid-documents/#aliasing-fields>

Might help bring the code base to a more familiar ruby footing??

irisfaraway: Good point. I'd definitely prefer more consistency with Ruby syntax so I'll make the change!

irisfaraway:
@Cruikshanks I've swapped as much over to snake_case as possible if you want to have another look. (Massive PR now, sorry!)

PR #7, opened 2017-12-06
Add transient object for renewals

irisfaraway:
https://eaflood.atlassian.net/browse/WC-125

We want to store information about a renewal as a user goes through the process, but we don't want to modify the existing registration until that process is complete. So we need an additional object to be stored in the database to capture this incomplete renewal.

You start the renewal process by going to /renew/regIdentifier (so for example /renew/CBDU1234).

At this point the app checks to make sure:

- the registration ID is in the correct format (eg. not an injection attack)
- the registration ID exists
- the registration isn't already being renewed

If any of these problems are detected, the user will get an error message informing them of the issue. These checks happen on both GET and POST requests.

If not, they will see a message saying "You are about to renew CBDU1234" and a button. If they click the button, a new TransientRegistration object is created in the database (although you can't do anything with it yet). The transient object copies all the data from the registration to itself so the user can begin to confirm or modify it (or at least will be able to in the future) without updating the registration itself.

Annoyingly the rebase against PR #6 has brought those changes in as well, resulting in an inflated "Files changed" count.


app/models/transient_registration.rb
```diff
@@ -0,0 +1,37 @@
...
+  validate :valid_reg_identifier?
+  validate :no_renewal_in_progress?
+  validate :registration_exists?
+
+  private
+
+  def valid_reg_identifier?
+    # Make sure the format of the reg_identifier is valid to prevent injection
+    # Format should be CBDU, followed by at least one digit
+    return unless reg_identifier.blank? || !reg_identifier.match?(/^CBDU[0-9]+$/)
```
Cruikshanks: Unfortunately we can't rely on all upper tier registrations starting with **CBDU**. Run the following query in MongoDb and you'll see we have Upper tier paid for registrations that start **CBDL**.

```javascript
db.registrations.find({
  $and: [
    { 'tier': { $exists: true } },
    { 'tier': 'UPPER'},
    { 'regIdentifier': /^CBDL/ }
  ]
})
```
So this will need to be updated to account for this.

spec/factories/transient_registration.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :transient_registration do
+    trait :has_required_data do
```
Cruikshanks: Might be worth adding a comment saying something like this is a placeholder/keeping the structure in . sync with the registration factory, though at this time the trait is not implemented.

app/forms/renewal_start_form.rb
```diff
@@ -0,0 +1,41 @@
...
+  def initialize(transient_registration)
+    @transient_registration = transient_registration
+    # Get values from transient registration so form will be pre-filled
+    self.reg_identifier = @transient_registration.reg_identifier
+  end
+
+  validates :reg_identifier, presence: true
+  validate :transient_registration_valid?
+
+  def submit(params)
```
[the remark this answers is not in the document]
irisfaraway: Good point. I've chosen to keep the first `params` as `params` as this will be more consistent with the other forms but have renamed the one getting data from the registration.

irisfaraway:
@Cruikshanks I've made the changes you suggested in the review 👍

PR #8, opened 2017-12-11
Set up 'happy path' for limited company renewals

irisfaraway:
https://eaflood.atlassian.net/browse/WC-124

The 'happy path' should cover the pages which the user will need to complete a renewal, provided they meet certain conditions (in this case, a limited company based in the UK).

This will be in a fairly skeletal form - just a heading, a back link and a continue button. It should provide a basis for testing and further development. Functionality like the address lookup and smart answers is not covered.

This will also handle implementing a state machine to manage how the user travels between pages.


irisfaraway:
Here's where this is at the moment:

- When you go to /renew/CBDU1234 (or whatever your reg_identifier is), a new transient object will be created, or if one is in progress already then it'll return to that one
- When you submit the form, if everything is valid, it goes to the next state and the next form
- When you click "back", it ignores whatever you've done on the form and goes to the previous state and form
- When you go away and come back to /renew/CBDU1234 (or any other page in the journey) you will be redirected to whichever form you were last on
- Only 3 pages exist so far - the start page, the business type and a stand-in for smart answers. Trying to "continue" past that point will result in errors because I haven't done the other pages yet.

I'm happy to crack on and do the rest of the pages in this form, BUT I am conscious that this is already a pretty big chunk of work and it might benefit from someone else having a look at it. So @Cruikshanks, would you mind casting an eye over this?

irisfaraway:
I've written a scaffolding script (adapted from the one for waste-permits) to generate the remaining forms and all their assorted files, because manually creating and editing 126 near-identical files did not sound like fun.

Until all the forms are added, there will always be two failing tests – these belong to the last form in the sequence so far, which expects to have something to proceed to and doesn't yet.

irisfaraway:
Ready for another review @Cruikshanks! Just a lot more of the same...

PR #15, opened 2017-12-21
Select type of business or organisation

irisfaraway:
https://eaflood.atlassian.net/browse/WC-90

As part of the renewal, users must state what type of business or organisation the registration belongs to. The form should have their existing choice pre-selected, but they should be able to change it (although changing the type may require making a new registration instead – we'll handle that in a later story).

Some options will not be pre-selected as the available options have changed slightly – for example, public body and local authority have been combined

This doesn't cover any routing that comes from selecting a different type. See [WC-91](https://eaflood.atlassian.net/browse/WC-91) for that.

### DIFF
--- a/config/locales/forms/business_type_forms/en.yml
+++ b/config/locales/forms/business_type_forms/en.yml
@@ -5,7 +5,7 @@ en:
       option_local_authority: Local authority or public body
       option_limited_company: Limited company (eg a registered company or plc)
       option_limited_liability_partnership: Limited liability partnership
-      option_individual: Individual (eg a householder)
+      option_individual: Individual (eg a sole trader)
       option_partnership: Partnership
       option_other: Other organisation (eg a trust, charity or club)
       option_overseas: Based overseas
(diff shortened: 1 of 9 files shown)


config/locales/forms/business_type_forms/en.yml
```diff
@@ -1,7 +1,15 @@
 en:
   business_type_forms:
     new:
-      heading: Business type
+      heading: What type of business or organisation are you?
+      option_local_authority: Local authority or public body
+      option_limited_company: Limited company (eg a registered company or plc)
+      option_limited_liability_partnership: Limited liability partnership
+      option_individual: Individual (eg a householder)
```
Cruikshanks: Small change. The content from this form came from FRAE which does deal with actual members of the public e.g. I have a house on the river and I want to make a change to the riverbank.

In WCR we don't ever deal with members of the public. So this needs to change to

```yaml
option_individual: Individual (eg a sole trader)
```

irisfaraway: Oops, good point. Should be fixed now.

PR #16, opened 2018-01-02
Use dotenv for environment variables

irisfaraway:
https://eaflood.atlassian.net/browse/WC-144

We should store some environment variables in a .env file. This is for security, consistency with other apps, and for ease of changing them.

We can load the variables using the dotenv gem.


.env.example
```diff
@@ -0,0 +1,41 @@
...
+
+# Settings for waste-carriers-frontend
+WCRS_FRONTEND_APP_DOMAIN="http://localhost:3000"
+
+# Application settings
+WCRS_REGISTRATION_EXPIRES_AFTER=3
+WCRS_REGISTRATION_RENEWAL_WINDOW=6
+WCRS_RENEWAL_CHARGE=105
+WCRS_TYPE_CHANGE_CHARGE=40
+WCRS_RENEWALS_DEVISE_SECRET_KEY=
```
Cruikshanks: If I've read the intent wrong, let me know. But it looks like rather than adding all the possible env vars at once, you have gone with adding only what is actually required by the renewals service at this time.

If this is the case I'd be consistent with that intent, and remove from the example things not currently required like this, and the google analytics stuff at the bottom.

irisfaraway: Good point. I've fixed that (and rebased).

PR #21, opened 2018-01-05
Report errors to Errbit

irisfaraway:
https://eaflood.atlassian.net/browse/WC-87

We should capture all exceptions and report them to Errbit, our Airbrake-complaint error tracking service.

We can test our Airbrake/Errbit setup by running `bundle exec rake airbrake:test` – this should send a test error report to our Errbit instance.


irisfaraway:
CodeClimate is mad because Airbrake is switched off when it runs the tests, so it never gets to run most of the initialiser file and then test coverage drops. @Cruikshanks Do we want Travis to be reporting to Errbit? I'm tempted to just eat the drop in test coverage on this one.

Cruikshanks:
Agreed @irisfaraway , we can eat it on this one. I'll get on with the review now unless you say otherwise.

irisfaraway:
@Cruikshanks I've added the Travis config and tested it out with an error (now rolled back) – it does indeed run in the test environment and so doesn't spam Errbit. So I think good to merge 👍

PR #24, opened 2018-01-08
Set up "skeleton" of happy path for Individuals

irisfaraway:
https://eaflood.atlassian.net/browse/WC-148

This story covers setting up a "skeleton" of the happy path for those who would select Individuals on the select org type page. Pages in the skeleton will mostly be blank – just a title and continue button.

This is so we can set up how users move between the pages, and also so we can start testing the process as a whole.

As we continue to work on the stories, these pages will be fleshed out with the proper forms.


irisfaraway:
@Cruikshanks Think you could have a look at this? If you're happy with this format I'll crack on with the other business types.

Copy needs work but I think we'll get to that when we actually make the pages. For now, we can at least differentiate between the different business types.

I did somehow muck up the `git mv` for two renamed files but the others should be OK 😅

config/locales/forms/registration_number_forms/en.yml
```diff
@@ -1,7 +1,8 @@
 en:
   registration_number_forms:
     new:
-      heading: What's the registration number of the company?
+      heading_limitedCompany: What's the registration number of the company?
+      heading_soleTrader: What's the registration number of the individual?
```
Cruikshanks: We should never have a need for this as an individual should never go down a path that requests a registration number.

I'd rather take this out so if we have something in the sole trader journey that relies on this translation existing we know about it.

On a side note, do we have the app configured to raise an exception in dev and test if a translation is missing?

db/seeds.rb
```diff
@@ -86,4 +87,79 @@
       confirmed: "no"
     }
   )
+
+  # Sole trader
```
Cruikshanks: Reckon this is fine for this PR, but we need to think about a different approach going forward especially as @timstone is now busy putting together his list of registrations he'd like to see seeded!

config/locales/forms/registration_number_forms/en.yml
irisfaraway: You're completely right. I was thinking it was weird that there was a reg number page for individuals but I must have misread the wireframe/journey map. Will sort this out!

And no, we don't have that configured – I'll look into it.

Cruikshanks: It's a little old, but I think it covers our version of rails https://robots.thoughtbot.com/foolproof-i18n-setup-in-rails

irisfaraway: Both these issues should now be fixed 👍

db/seeds.rb
irisfaraway: Agreed, this could get messy pretty quickly – should we have a story to look into how to handle seeding in a sustainable way?

Cruikshanks: Agreed and done! Implement sustainable method for seeding the database

PR #27, opened 2018-01-09
Set up "skeleton" of happy path for "other" organisations

irisfaraway:
https://eaflood.atlassian.net/browse/WC-150

This story covers setting up a "skeleton" of the happy path for those who would select Other organisations on the select org type page. The "other" type covers charities, trusts, clubs and other organisations which should be lower tier registrations. For this reason, selecting "other" as your business type shoud tell you to start a new lower tier registration instead of renewing.

### DIFF
--- a/config/locales/forms/cannot_renew_type_change_forms/en.yml
+++ b/config/locales/forms/cannot_renew_type_change_forms/en.yml
@@ -1,15 +1,15 @@
 en:
-  cannot_renew_lower_tier_forms:
+  cannot_renew_type_change_forms:
     new:
-      heading: You should not renew %{reg_identifier}
-      paragraph_1: Based on the answers to our questions, you are eligible to register as a lower tier waste carrier.
-      paragraph_2: There is no charge to registered as a lower tier waste carrier, but you will need to create a new registration.
+      heading: You cannot renew %{reg_identifier}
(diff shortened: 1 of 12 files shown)


irisfaraway:
This will require some more logic when users can be routed to lower tier for other reasons (eg smart answers) but we'll cross that bridge when we come to it 🙂

config/locales/forms/cannot_renew_type_change_forms/en.yml
```diff
@@ -1,15 +1,15 @@
 en:
-  cannot_renew_type_change_forms:
+  cannot_renew_lower_tier_forms:
     new:
-      heading: You cannot renew %{reg_identifier}
-      paragraph_1: You've told us that your business type has changed since you originally registered.
-      paragraph_2: Because of this change, you cannot renew %{reg_identifier} and instead must make a new registration.
+      heading: You should not renew %{reg_identifier}
+      paragraph_1: Based on the answers to our questions, you are eligible to register as a lower tier waste carrier.
+      paragraph_2: There is no charge to registered as a lower tier waste carrier, but you will need to create a new registration.
```
Cruikshanks: Typo 😁 !

`There is no charge to registered as a ...` should be `There is no charge to register as a ...`

irisfaraway:
@Cruikshanks Typo sorted 👍

PR #28, opened 2018-01-09
Set up "skeleton" of happy path for partnerships

irisfaraway:
https://eaflood.atlassian.net/browse/WC-149

This story covers setting up a "skeleton" of the happy path for those who would select Partnership on the select org type page. Pages in the skeleton will mostly be blank – just a title and continue button.

This is so we can set up how users move between the pages, and also so we can start testing the process as a whole.

As we continue to work on the stories, these pages will be fleshed out with the proper forms.


irisfaraway:
Will need a rebase after https://github.com/DEFRA/waste-carriers-renewals/pull/27 merges.

PR #31, opened 2018-01-10
Improve database seeding

irisfaraway:
https://eaflood.atlassian.net/browse/WC-167

We need some seed data for the database for the dev environment and for running acceptance tests against. Currently this is handled in a single file – however, this is becoming messy to manage. We need to find a more sustainable way to handle this.

WIP.


irisfaraway:
Closing this for now, as after looking into implementation, there are some other stories we need to do first.

PR #32, opened 2018-01-10
Add validation to Registrations to ensure regIdentifier is unique

irisfaraway:
https://eaflood.atlassian.net/browse/WC-169

The registration number (known as the regIdentifier in the model) is expected to be unique. Validation is in place for TransientRegistrations. This change adds validation for the Registration model as well.


irisfaraway:
@Cruikshanks Ready for review! (Although you may need to give that CodeClimate build a kick – think it's stuck and I don't have permission to restart it 😬 )

PR #34, opened 2018-01-11
Implement reg_identifier number generator

irisfaraway:
https://eaflood.atlassian.net/browse/WC-170

Registrations should have reg_identifiers in the format of CBDU or CBDL, followed by a sequential number (1, 2, 3...). These numbers need to be unique. The renewals service will generally be dealing with existing registrations which already have a number, but we also need to be able to seed new registrations for testing purposes, and these should not clash with any registrations which were created through the usual process.

This change adds reg_identifier generation which will play nicely with the existing service.


irisfaraway:
@Cruikshanks This is ready for a review. Let me know if you have any questions!

PR #37, opened 2018-01-15
Improve database seeding

irisfaraway:
https://eaflood.atlassian.net/browse/WC-167

We need some seed data for the database for the dev environment and for running acceptance tests against. This change switches from loading all seeds in a single script to storing seeds as multiple JSON files and then loading them. This approach should make it easier to add and modify seeds.


irisfaraway:
@Cruikshanks This is ready for a review. You'll probably want to play around with the seed data, especially adding different `regIdentifier`s to change the order in which seeds are added.

Hakiri has flagged up an issue – I've looked into this and don't think it is a risk as the value it's concerned about (the regid counter) should only be modified by the application and not by any data from users, etc. What do you think?

irisfaraway:
We've also got some docs (🎉 ) here: https://github.com/DEFRA/waste-carriers-renewals/wiki/Database-seeding

PR #38, opened 2018-01-15
Implement smart answers

irisfaraway:
https://eaflood.atlassian.net/browse/WC-92

This change adds the "smart answers" forms to help determine if a user is an upper tier or lower tier waste carrier.

WIP


irisfaraway:
@Cruikshanks Think this is ready for a review! Apologies for the mega-PR... some of this is down to a form objects refactor prompted by CodeClimate, but mostly it's just because replacing 1 form with 4 adds a lot of new files. Happy to run through what I've done in person if you'd like.

This doesn't cover informing the user if they should switch to a lower-tier registration (WC-94) – I'll pick that up next.

PR #40, opened 2018-01-18
Revert "Bundle Update on 2018-01-18"

irisfaraway:
Reverts DEFRA/waste-carriers-renewals#39


irisfaraway:
Reverting as this drops support for Mongo 2.4, which we are using for the older parts of the service.

PR #42, opened 2018-01-18
Changing from upper to lower tier exits renewals process

irisfaraway:
https://eaflood.atlassian.net/browse/WC-94

If the user's responses to the 'smart answers' section of the service indicate that they should have a lower tier registration, they should be informed of this and directed to make a new, free, lower tier registration instead of renewing.


irisfaraway:
@Cruikshanks That photo is great!

PR #44, opened 2018-01-19
Refactor model tests for Transient Registration

irisfaraway:
This file has become bloated and repetitive, mostly due to the bulk of the state change tests. This change refactors it to remove duplication and split it into separate files when necessary.

WIP.


irisfaraway:
@Cruikshanks Still plenty of work to do on the refactoring front, but I think this is reviewable now. Basically just splits up `transient_registration_spec.rb` into separate files for each workflow state. I also tweaked the `business_type_form` one so it uses a single permutation table to make sure it's producing the right result instead of ~200 lines of near-identical tests.

PR #46, opened 2018-01-22
Users can select carrier/broker/dealer type

irisfaraway:
https://eaflood.atlassian.net/browse/WC-95

After smart answers, we should ask users what their carrier/broker/dealer status is. Changing this from what was originally selected when they first registered does result in an additional fee, although calculating that is not part of this story.


irisfaraway:
@Cruikshanks This was a quick one! Think it's ready for a review.

PR #47, opened 2018-01-22
Renewal information page

irisfaraway:
https://eaflood.atlassian.net/browse/WC-96

After completing smart answers and answering "Do you carry the waste yourselves, or arrange for others to do it?" we can give users more information about their renewal and how much it will cost.


irisfaraway:
@Cruikshanks Here's another one for you to have a look at (but no rush) – must be speedy story day.

PR #49, opened 2018-01-23
Get the registration number of a limited company or LLP

irisfaraway:
https://eaflood.atlassian.net/browse/WC-97

Limited companies and LLPs must provide their registration number for Companies House. We then need to validate that the number is in the correct format, then query Companies House to make sure the company/LLP is active. We should then inform the user if Companies House says there's a problem, or return the details from Companies House so the user can confirm all is well.


irisfaraway:
@Cruikshanks I think this is ready for a review now!

I ended up not using the ea-validation gem (much as I wanted to) as I only wanted to make requests to Companies House if the other `company_no` validations had passed, and that was proving tricky to get working when those validations were happening in a separate gem. If you can think of a clever way to do both then let me know.

app/models/companies_house_caller.rb
```diff
@@ -0,0 +1,45 @@
+class CompaniesHouseCaller
```
Cruikshanks: I know this is where this logic has sat in other projects, but I've never been comfortable with it. IMHO I think a model should represent something that persists, and so far we've followed this pattern.

Therefore I'd like to suggest moving this to being a service object, or into the `lib/` folder.

spec/forms/registration_number_forms_spec.rb
```diff
@@ -4,40 +4,44 @@
...
 
+    describe "#reg_identifier" do
       context "when a reg_identifier meets the requirements" do
         before(:each) do
           registration_number_form.reg_identifier = transient_registration.reg_identifier
         end
 
         it "is valid" do
-          expect(registration_number_form).to be_valid
+          VCR.use_cassette("registration_number_form_valid_company_no", re_record_interval: 14.days) do
```
Cruikshanks: VCR supposedly has the ability to apply [default cassette options](https://relishapp.com/vcr/vcr/v/3-0-3/docs/configuration/default-cassette-options) at the config level.

Rather than having to specify it in every call, I'd like to see if we can try and set it at the global level.

spec/forms/registration_number_forms_spec.rb
```diff
@@ -46,17 +50,74 @@
...
 
+        it "is not valid" do
+          VCR.use_cassette("registration_number_form_valid_company_no", re_record_interval: 14.days) do
+            expect(registration_number_form).to_not be_valid
+          end
+        end
+      end
+    end
+
+    describe "#company_no" do
```
Cruikshanks: After your opinion on this one. Should we / do we need also to have a test to cover someone entering a reg number as `1709418` rather than `01709418`?

irisfaraway: Yeah, I think this would be good to add.

spec/forms/registration_number_forms_spec.rb
irisfaraway: Sounds good to me – I must have ended up on the wrong page of the docs. Will add this in!

app/models/companies_house_caller.rb
irisfaraway: It felt a bit funny to me as well so I'm happy to move it 👍

irisfaraway:
@Cruikshanks I've made the changes you suggested, so this is ready for another look when you're free!

PR #55, opened 2018-01-29
 Get the name of the company/organisation

irisfaraway:
https://eaflood.atlassian.net/browse/WC-99

This is the trading name, rather than the registered name.


irisfaraway:
@Cruikshanks Thought I'd do a nice easy one for Monday morning 😅

PR #57, opened 2018-01-29
Changing company's registration number exits renewals process

irisfaraway:
https://eaflood.atlassian.net/browse/WC-98

If the company's registration number changes, it is considered a new entity. This means the registration cannot be renewed and they must make a new one instead.

### DIFF
--- a/app/models/concerns/can_change_workflow_status.rb
+++ b/app/models/concerns/can_change_workflow_status.rb
@@ -304,6 +304,9 @@ module CanChangeWorkflowStatus
 
   def require_new_registration_based_on_company_no?
     old_company_no = Registration.where(reg_identifier: reg_identifier).first.company_no.to_s
+    # It was previously valid to have company_nos with less than 8 digits
+    # The form prepends 0s to make up the length, so we should do this for the old number to match
+    old_company_no = "0#{old_company_no}" while old_company_no.length < 8
     old_company_no != company_no
   end
 
--- a/app/models/concerns/can_change_workflow_status.rb
+++ b/app/models/concerns/can_change_workflow_status.rb
@@ -108,6 +108,10 @@ module CanChangeWorkflowStatus
                     to: :registration_number_form
 
         transitions from: :registration_number_form,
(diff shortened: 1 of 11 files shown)


irisfaraway:
@Cruikshanks The check to see if the `company_no` has changed needs some leniency, especially if the validations here are different to the ones in waste-carriers-frontend. I had a look at the existing validations and I think the only difference is that number-only strings with fewer than 8 digits were previously allowed, whereas now we add 0s to the start. I think I've covered this but let me know if you think of anything else.

app/models/concerns/can_change_workflow_status.rb
```diff
@@ -295,6 +302,14 @@ def switch_to_lower_tier_based_on_smart_answers?
     false
   end
 
+  def require_new_registration_based_on_company_no?
+    old_company_no = Registration.where(reg_identifier: reg_identifier).first.company_no.to_s
+    # It was previously valid to have company_nos with less than 8 digits
```
Cruikshanks: What do you think of making this correction to the company number when we convert the registration to the transient registration object?

You can then revert back to the simpler logic here, which is more explicit about what we were actually testing for.

💡 💭 😐 ❓

irisfaraway: Ooh, I like this idea! Will make the change.

irisfaraway: @Cruikshanks Just to clarify – do you mean modifying the number on the original registration itself? Would this have any knock-on effects for the other apps which I should be aware of?

Cruikshanks: Err, no. I meant as we migrate the data from  registration to  the transient we correct it.

However now thinking I need to go back to the original version of the check you put in to see if you were doing `param vs transient` or `transient vs registration`.

If its the latter, then I don't think my suggestion is valid.

irisfaraway: It was the latter, unfortunately, as entering a different company number is valid (if the number itself is OK) – it just means you should exit the renewals process.

PR #59, opened 2018-01-29
updating companies house number to valid number

timstone:
Updated companies house numbers to a valid company, also removed value for non limited, and LLP companies.

### DIFF
--- a/db/seeds/CBDU1.json
+++ b/db/seeds/CBDU1.json
@@ -6,7 +6,7 @@
   "isMainService": "yes",
   "onlyAMF": "no",
   "companyName": "Limited Company Seed",
-  "companyNo": 1234567,
+  "companyNo": "946107",
   "firstName": "Test",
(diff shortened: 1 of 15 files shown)


db/seeds/CBDU1.json
```diff
@@ -6,7 +6,7 @@
   "isMainService": "yes",
   "onlyAMF": "no",
   "companyName": "Limited Company Seed",
-  "companyNo": 1234567,
+  "companyNo": "946107",
```
irisfaraway: This is `946107` whereas the others are `00946107` – is this intentional?

Might come in handy for testing the routing to new registrations due to changing the Companies House number (eg `946107` to `00946107` should be a valid change), but I don't know if you'd want a separate seed for that.

timstone: Yes thought I'd throw one in without the leading zeros, so it should catch any issues with that.

PR #61, opened 2018-01-30
 Add terms and conditions for Ordnance Survey

irisfaraway:
https://eaflood.atlassian.net/browse/WC-193

We use the Ordnance Survey's address data to look up and fill in addresses based on a user-provided postcode. This means we need to include certain terms and conditions for its use.


irisfaraway:
@Cruikshanks Hopefully not much to look at here!

PR #62, opened 2018-01-30
Search for address by postcode

irisfaraway:
https://eaflood.atlassian.net/browse/WC-100

Users should be able to get their address by searching with the postcode. This saves time typing in the address, but also ensures data that is more compliant with our standards. We use OS Places to provide this data.

WIP


irisfaraway:
They were green on my machine 😭

irisfaraway:
@Cruikshanks After 8 million years, and roughly as many commits, I think this is finally there. I know this is a big one so happy to talk through it in person or over the phone first.

app/services/address_finder_service.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    begin
+      response = RestClient::Request.execute(
+        method: :get,
+        url: @url
+      )
+
+      JSON.parse(response)
+    rescue RestClient::BadRequest
+      Rails.logger.debug "OS Places: resource not found"
```
Cruikshanks: If OS places was down, would we fall into this rescue block? In which case it needs to send a notification toi Airbrake as well.

Cruikshanks: I did find this example https://stackoverflow.com/questions/25552239/webmock-simulate-failing-api-no-internet-timeout but I think its one for another time hence marked PR as approved.

PR #74, opened 2018-02-13
Provide an address manually

irisfaraway:
https://eaflood.atlassian.net/browse/WC-101

While we prefer users to add addresses using the OS Places API, some users will not be able to use it, for example if they are overseas, if their address is not listed or if the service is down. These users should be able to fill in a separate form with details of their address.


irisfaraway:
@Cruikshanks It's been a journey... happy to talk this through (more) in person. 😅

app/models/concerns/can_have_registration_attributes.rb
```diff
@@ -37,5 +37,15 @@ module CanHaveRegistrationAttributes
     field :declaration,                                     type: Integer # Unsure of type
     field :regIdentifier, as: :reg_identifier,              type: String
     field :expires_on,                                      type: DateTime
+
+    def contact_address
```
Cruikshanks: Nice helpers! 👍

app/services/address_finder_service.rb
```diff
@@ -20,7 +20,11 @@ def search_by_postcode
       :not_found
     rescue RestClient::ExceptionWithResponse => e
       Airbrake.notify(e)
-      Rails.logger.error "Os Places error: " + e.to_s
+      Rails.logger.error "OS Places error: " + e.to_s
```
Cruikshanks: Do you think there is any value tweaking the message for these errors to differentiate them in the log e.g. `OS Places response error` and `OS Places socket error`?

config/locales/forms/company_address_manual_forms/en.yml
```diff
@@ -0,0 +1,43 @@
+en:
+  company_address_manual_forms:
+    new:
+      heading_localAuthority: What's the address of the local authority or public body?
+      heading_limitedCompany: What's the address of the company?
+      heading_limitedLiabilityPartnership: What's the address of the limited liability partnership?
+      heading_partnership: What's the address of the partnership?
+      heading_soleTrader: What's the address of the individual?
```
Cruikshanks: Needs to be *What's the address of the business?*

app/services/address_finder_service.rb
irisfaraway: Sounds like a good idea to me. I'll tweak it now.

irisfaraway: @Cruikshanks Fixed!

PR #88, opened 2018-02-26
 Generate relative dates for seed files

irisfaraway:
When seeding registrations, we sometimes want their dates to be relative to the current date (eg 3 months ago) as opposed to a specific date (eg 26 February 2018). This change lets us specify relative dates in seed data and parse the seed data appropriately.

When we parse each JSON seed, check to see if the JSON's `date_flag` is set and then generate all time-sensitive dates based on the flag. Options are expired, in the renewal window or outside the renewal window.

If no `date_flag` is set, the script will load whichever dates are provided in the seed data.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,10 +5,17 @@
 # remains in that state without having to manually update the dates.
 
 def registered_date(flag)
+  # Resetting variables here so it's easier to read calculations below
+  expires_after = Rails.configuration.expires_after
+  renewal_window = Rails.configuration.renewal_window
+
   dates = {
-    expired: 37.months.ago,
-    in_renewal_window: 35.months.ago,
-    outside_renewal_window: 12.months.ago
+    # Registration should have expired 1 month ago
+    expired: expires_after.years.ago - 1.month,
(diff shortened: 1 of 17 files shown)


db/seeds.rb
```diff
@@ -1,10 +1,29 @@
...
-#   Mayor.create(name: 'Emanuel', city: cities.first)
+# This file adds seeds by parsing the JSON files in the db/seeds/ folder.
+
+# If a "date_flag" is set for a seed, we generate the dates dynamically.
+# This ensures a seed which is always intended to be in a time-sensitive state
+# remains in that state without having to manually update the dates.
+
+def registered_date(flag)
+  dates = {
+    expired: 37.months.ago,
```
Cruikshanks: Should these not be making use of the config values for expiry and renewal window to work out how many months to use? Or have I missed something?

irisfaraway: @Cruikshanks They should and that makes complete sense. I'll tweak that now.

irisfaraway: @Cruikshanks Should be sorted now. It's a little awkward to read but hopefully the comments clear it up.

PR #89, opened 2018-02-26
Add a key person

irisfaraway:
https://eaflood.atlassian.net/browse/WC-105

All business types have different sorts of key people whose details are required for the registration. Most can have multiple key people, but the scope of this story just covers adding one.

### DIFF
--- a/app/forms/key_people_form.rb
+++ b/app/forms/key_people_form.rb
@@ -41,7 +41,7 @@ class KeyPeopleForm < BaseForm
   end
 
   def old_enough?
-    return unless date_of_birth.present?
+    return false unless date_of_birth.present?
 
     age_limits = {
       limitedCompany: 16.years,
@@ -57,5 +57,6 @@ class KeyPeopleForm < BaseForm
 
     error_message = "age_limit_#{business_type}".to_sym
     errors.add(:date_of_birth, error_message)
+    false
   end
 end
--- a/app/forms/key_people_form.rb
+++ b/app/forms/key_people_form.rb
@@ -27,7 +27,7 @@ class KeyPeopleForm < BaseForm
   validates :first_name, presence: true, length: { maximum: 35 }
   validates :last_name, presence: true, length: { maximum: 35 }
   validate :old_enough?
-  validates_with DobValidator
+  validates_with DateOfBirthValidator
 
   private
 
--- a/app/forms/key_people_form.rb
+++ b/app/forms/key_people_form.rb
@@ -1,6 +1,6 @@
 class KeyPeopleForm < BaseForm
   attr_accessor :business_type
-  attr_accessor :first_name, :last_name, :dob_day, :dob_month, :dob_year
+  attr_accessor :first_name, :last_name, :dob_day, :dob_month, :dob_year, :date_of_birth
 
   def initialize(transient_registration)
     super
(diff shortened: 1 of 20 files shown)


app/forms/key_people_form.rb
```diff
@@ -9,10 +10,52 @@ def initialize(transient_registration)
...
+  def add_key_person
+    KeyPerson.new(first_name: first_name,
+                  last_name: last_name,
+                  dob_day: dob_day,
+                  dob_month: dob_month,
+                  dob_year: dob_year,
+                  person_type: "key")
+  end
+
+  def old_enough?
```
Cruikshanks: It's my belief that methods phrased as questions should only return `true`, `false` or throw an exception.

So I would decide what result you want to return in the case of `date_of_birth` not being present, and ensure false is returned if `date_of_birth < age_cutoff_date`

app/validators/date_of_birth_validator.rb
```diff
@@ -0,0 +1,48 @@
...
+  def validate_field(record, type, field)
+    return unless field_present?(record, type, field)
+    return unless field_is_integer?(record, type, field)
+    field_is_in_correct_range?(record, type, field)
+  end
+
+  def field_present?(record, type, field)
+    return true unless field.blank?
+    error_message = "#{type}_blank".to_sym
+    record.errors.add(:date_of_birth, error_message)
```
Cruikshanks: Similar to my previous comment, if the field is not present this method should return `false`.

app/validators/date_of_birth_validator.rb
```diff
@@ -0,0 +1,48 @@
...
+  def field_is_in_correct_range?(record, type, field)
+    ranges = {
+      day: 1..31,
+      month: 1..12,
+      year: 1900..(Date.today.year.to_i)
+    }
+
+    return true if ranges[type].include?(field)
+    error_message = "#{type}_range".to_sym
+    record.errors.add(:date_of_birth, error_message)
```
Cruikshanks: Similar to my previous comment, if the field is not present this method should return `false`.

app/validators/date_of_birth_validator.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def dob_is_a_date?(record)
+    return true if record.date_of_birth.is_a? Date
+    record.errors.add(:date_of_birth, :not_a_date)
```
Cruikshanks: Similar to my previous comment, if the field is not present this method should return `false`.

config/locales/forms/key_people_forms/en.yml
```diff
@@ -6,14 +6,51 @@ en:
...
-      heading_soleTrader: Individual details
-      error_heading: Something is wrong
+      heading_soleTrader: Business owner details
+      description_1_localAuthority: Provide the name and date of birth for each chief executive of this local authority or public body.
+      description_1_limitedCompany: Provide the name and date of birth for each director of this business.
+      description_1_limitedLiabilityPartnership: Provide the name and date of birth for each partner of this business.
+      description_1_overseas: Provide the name and date of birth for each owner of this business or organisation.
+      description_1_partnership: Provide the name and date of birth for each partner of this business.
+      description_1_soleTrader: Provide the name and date of birth for the owner of this business.
+      description_2: Add their details one person at a time. If you want to add more people, select 'add another person'.
```
Cruikshanks: Suggest it should be

> If you need to add more people ...

rather than

> If you want to add more people

app/forms/key_people_form.rb
```diff
@@ -9,10 +10,52 @@ def initialize(transient_registration)
...
+    return unless date_of_birth.present?
+
+    age_limits = {
+      limitedCompany: 16.years,
+      limitedLiabilityPartnership: 17.years,
```
Cruikshanks: Fine for now but have you asked whether LLP's should have the same age restriction as Limited companies?

spec/forms/key_people_forms_spec.rb
```diff
@@ -21,36 +28,242 @@
...
+    end
+
+    describe "#date_of_birth" do
+      context "when a date_of_birth meets the requirements" do
+        it "is valid" do
+          expect(key_people_form).to be_valid
+        end
+      end
+
+      context "when a date of birth is not a valid date" do
```
Cruikshanks: Think we should also have a test for a date that is invalid e.g. 31-02-2000

spec/forms/key_people_forms_spec.rb
```diff
@@ -21,36 +28,242 @@
...
+          expect(key_people_form).to_not be_valid
+        end
+      end
+
+      shared_examples_for "age limits for key people" do |business_type, age_limit|
+        before(:each) do
+          key_people_form.business_type = business_type
+        end
+
+        it "should not be valid when at the age limit" do
```
Cruikshanks: Am I not getting this? The description is *should not be valid when at the age limit* but the expectation is checking that it is valid.

irisfaraway: Nah, this is just me copying and pasting 😅

app/forms/key_people_form.rb
irisfaraway: I haven't, but this is a good point. I'll find out.

spec/forms/key_people_forms_spec.rb
irisfaraway: Added some tests against the KeyPerson model to cover this.

app/forms/key_people_form.rb
irisfaraway: This should be fixed now.

irisfaraway:
@Cruikshanks I think I've addressed all your comments (except the LLP age thing – we're looking into it!)

PR #90, opened 2018-02-28
changing expiry flag for seed data

timstone:
Changing the flag status for two registrations so that it matches my tests

### DIFF
--- a/db/seeds/CBDU17.json
+++ b/db/seeds/CBDU17.json
@@ -54,7 +54,7 @@
   "regIdentifier": "CBDU17",
   "metaData": {
     "anotherString": "userDetailAddedAtRegistration",
-    "status": "ACTIVE",
+    "status": "EXPIRED",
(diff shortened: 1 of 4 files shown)


db/seeds/CBDU17.json
```diff
@@ -0,0 +1,65 @@
...
+      "confirmed": "no"
+    }
+  }],
+  "accountEmail": "user@waste-exemplar.gov.uk",
+  "declaredConvictions": "no",
+  "declaration": 1,
+  "regIdentifier": "CBDU17",
+  "metaData": {
+    "anotherString": "userDetailAddedAtRegistration",
+    "status": "ACTIVE",
```
irisfaraway: Should the status be `EXPIRED` here or are we just testing the date? Ditto for the other expired seed.

timstone: Good spot, I've updated

PR #92, opened 2018-03-01
Configure Tidelift to not check Rails version

irisfaraway:
We are sticking with Rails 4 for the time being, so this config change should stop PRs from getting blocked because newer versions are available.


irisfaraway:
Unfortunately we can't find a way to disable the security checks which are causing the Rails 4 issue, so disabling Tidelift for now and closing this branch. If we find a solution in the future, we'll re-enable it.

PR #95, opened 2018-03-05
Tweak key people validation handling and heading

irisfaraway:
These changes are following the QA review.

The bulk of the changes are about how we handle validation messages. This commit reduces the number of validation errors which a user can get at one time:

- Previously a totally blank form would trigger about 6 different messages - now it should just be one
- Date validation errors have also been simplified so if all fields are missing, we only display a single message
- We also only check that the complete date is a valid date once we've validated the individual fields to reduce the number of errors users get
- Date fields which don't contain valid integers are no longer converted to 0, which gives us better messages too

Also changed the heading text for overseas users to match the wireframe.

### DIFF
--- a/app/validators/date_of_birth_validator.rb
+++ b/app/validators/date_of_birth_validator.rb
@@ -1,19 +1,34 @@
 class DateOfBirthValidator < ActiveModel::Validator
   def validate(record)
+    # Make sure we have any date fields to validate before proceeding
+    return false if all_fields_empty?(record)
+
+    # Validate the content of individual fields
     fields = { day: record.dob_day, month: record.dob_month, year: record.dob_year }
 
+    all_fields_valid = true
     fields.each do |type, field|
-      validate_field(record, type, field)
+      next if field_is_valid?(record, type, field)
+      all_fields_valid = false
     end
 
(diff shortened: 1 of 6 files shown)


irisfaraway:
@Cruikshanks Would you mind having a look at this? I had to make a fair amount of changes to how this form's validations work under the hood and it might benefit from some second eyes!

irisfaraway:
@Cruikshanks Would you mind having a quick look at this when you've got the time? I had to make some fairly big changes to how this form is validated (added another custom validator, which then calls a second custom validator) so it could probably benefit from a second pair of eyes.

app/validators/date_of_birth_validator.rb
```diff
@@ -1,19 +1,39 @@
...
   end
 
   private
 
-  def validate_field(record, type, field)
-    return unless field_present?(record, type, field)
-    return unless field_is_integer?(record, type, field)
+  def all_fields_empty?(record)
+    fields = [record.dob_day, record.dob_month, record.dob_year].compact
+    return false if fields.any?
```
Cruikshanks: 🍬 👍

PR #97, opened 2018-03-12
Check which country users are based in

irisfaraway:
https://eaflood.atlassian.net/browse/WC-238

This PR adds a new form to ask where the business is based - England, Wales, Scotland, Northern Ireland or overseas.

Wales, Scotland and Northern Ireland have their own registers for waste carriers. This form will allow us to direct those users to the correct service which they should be registering with. Users can continue to register with the EA if they want, but this will at least make them aware of the other registry.

Overseas users operating in England should register with the EA - however, there are some differences in what forms they see (for example, the address forms). Currently we filter them out using the 'overseas' option on the business type form, but 'overseas' makes a lot more sense as a location than a business type.

### DIFF
--- a/app/forms/location_form.rb
+++ b/app/forms/location_form.rb
@@ -11,6 +11,9 @@ class LocationForm < BaseForm
     self.location = params[:location]
     attributes = { location: location }
 
+    # Set the business type to overseas when required as we use this for microcopy
+    attributes[:business_type] = "overseas" if location == "overseas"
+
     super(attributes, params[:reg_identifier])
   end
 
--- a/app/forms/location_form.rb
+++ b/app/forms/location_form.rb
@@ -13,4 +13,10 @@ class LocationForm < BaseForm
 
     super(attributes, params[:reg_identifier])
   end
+
+  validates :location, inclusion: { in: %w[england
+                                           northern_ireland
+                                           scotland
+                                           wales
+                                           overseas] }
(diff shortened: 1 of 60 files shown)


irisfaraway:
@Cruikshanks Ready for a review! Unfortunately adding extra pages means lots of new files. Probably more manageable to go through commit-by-commit. Happy to go through it in person too.

app/forms/location_form.rb
```diff
@@ -0,0 +1,25 @@
...
+    self.location = params[:location]
+    attributes = { location: location }
+
+    # Set the business type to overseas when required as we use this for microcopy
+    attributes[:business_type] = "overseas" if location == "overseas"
+
+    super(attributes, params[:reg_identifier])
+  end
+
+  validates :location, inclusion: { in: %w[england
```
Cruikshanks: My one pondering is whether there is a call to centralise lists such as this (business type being the other example) but as yet can't point at a specific reason why. So more a question of whether you've had any similar thoughts.

irisfaraway: @Cruikshanks This had crossed my mind as well. I guess one benefit would be that when we have a form with radio buttons for each of these options, we could iterate over the existing list rather than adding them all manually. What do you think? Shouldn't take too long to implement, although I'd probably do it as a separate PR if we're doing business types too.

PR #103, opened 2018-03-16
added new partner seed

timstone:
Adding new partner seed for acceptance testing


irisfaraway:
Have re-run scan on Hakiri and it's fine, even if it hasn't reported back to GitHub.

PR #104, opened 2018-03-16
Users can add people with relevant convictions

irisfaraway:
https://eaflood.atlassian.net/browse/WC-108

If a user answers yes to "Has anyone involved in managing your organisation been convicted of an environmental offence in the past 12 months?" they should be prompted to provide the details of the person with the offence (but not the details of the offence).

This should be a blank form as any data from the previous registration is out of date.

WIP


irisfaraway:
The drop in test coverage is because we're not testing PersonForm on its own, and so never call the NotImplementedErrors. I don't see much point in testing this but happy to add some if you disagree @Cruikshanks

Cruikshanks:
Oh, and just to confirm agree with not worrying about the drop in test coverage.

PR #110, opened 2018-04-03
Enter the name of the person to contact

irisfaraway:
https://eaflood.atlassian.net/browse/WC-109

"Who should we contact about this registration?" Allow users to provide the name of a contact person for their registration and renewal. Fields should be pre-populated with the name the user previously entered.


irisfaraway:
@Cruikshanks I think this is ready to review!

PR #115, opened 2018-04-06
Add contact telephone number

irisfaraway:
https://eaflood.atlassian.net/browse/WC-110

This change adds the form for "What's the telephone number of the person we should contact?"


irisfaraway:
Waiting for https://github.com/DEFRA/waste-carriers-renewals/pull/112 to merge before cracking on with validation for this one.

spec/forms/contact_phone_forms_spec.rb
```diff
@@ -13,37 +18,90 @@
...
+        before(:each) do
+          contact_phone_form.phone_number = "01234 567 890 123"
+        end
+
+        it "is not valid" do
+          expect(contact_phone_form).to_not be_valid
+        end
+      end
+
+      context "when a phone_number is not a valid number" do
```
Cruikshanks: Because of how 'clever' the phonelib library can be, and coming at specs as a form of documentation, I would suggest being a little clearer with the contexts e.g.

```ruby
context "when phone_number is is not a number" do
  before(:each) do
    contact_phone_form.phone_number = "foo"
  end
end
context "when phone_number is not a valid number" do
  before(:each) do
    # It might look valid, but actually this is not a recognised number
    contact_phone_form.phone_number = "0117 785 3149"
  end
end
```

spec/requests/contact_phone_forms_spec.rb
```diff
@@ -56,12 +56,14 @@
         context "when valid params are submitted" do
           let(:valid_params) {
             {
-              reg_identifier: transient_registration[:reg_identifier]
+              reg_identifier: transient_registration[:reg_identifier],
+              phone_number: "01234 567890"
```
Cruikshanks: Does this work (using 01234 567890)?

When I try this on the WEX service I get "This telephone number is not recognised"

irisfaraway: It validates for me – that's what we've been using for the seed data as well.

spec/forms/contact_phone_forms_spec.rb
irisfaraway: Sounds like a good idea to me! A good way to illustrate what the gem is actually doing. Will make changes now.

spec/requests/contact_phone_forms_spec.rb
Cruikshanks: 🤦‍♂️ 
screen shot 2018-04-10 at 09 35 40

...or "why it may be wise to not review PR's late at night!"

irisfaraway: Hahaha, we've all been there...

PR #117, opened 2018-04-09
 Search for a contact address by UK postcode

irisfaraway:
https://eaflood.atlassian.net/browse/WC-112

"What's the address of the person we should contact?" Add forms to allow UK-based users to enter a postcode so they can search for a contact address. Selection and saving of the contact address will be added in a separate PR to keep things manageable.

This should function the same way as the registered address forms earlier in the process.

Non-UK users or users who choose to enter their addresses manually will see a separate form (to be worked on as a separate feature).


irisfaraway:
Cheers @Cruikshanks. It's weirdly satisfying to implement too!

PR #118, opened 2018-04-09
Enter contact address manually

irisfaraway:
https://eaflood.atlassian.net/browse/WC-113

Overseas users, or users who can't use the postcode lookup for whatever reason, should have the opportunity to enter their address manually.

This should behave the same way as the manual address entry form for the registered address.

Because the postcode page has been set up in a separate branch (https://github.com/DEFRA/waste-carriers-renewals/pull/117) references to the contact_postcode_form have been removed for now. So this is probably best tested with an overseas registration if you want a realistic representation of the form in the workflow.


Cruikshanks:
I've had a look, and am happy to approve, but I think they'll be some changes once you merge #117 that are best just done here, which will then invalidate that approval.

Let me know when merged and this branch rebased, and I'll come back and approve then.

irisfaraway:
@Cruikshanks That makes sense. I've merged in the postcode bits, so this should be ready for a review now.

PR #119, opened 2018-04-09
Save the position of relevant conviction people

irisfaraway:
This change fixes a bug which meant that we weren't saving the 'position' attribute when we created a new person with a relevant conviction.


irisfaraway:
@Cruikshanks Just something I noticed while working on a different story! Oops.

PR #125, opened 2018-04-13
Rescue SocketErrors for Companies House lookup

irisfaraway:
When we send a company no to Companies House for validation, we should rescue any SocketErrors we get back and present a nicely formatted error instead. This still blocks progression but at least it doesn't show a scary error page.

(Ran into this one while working offline)

### DIFF
--- a/app/services/companies_house_service.rb
+++ b/app/services/companies_house_service.rb
@@ -26,6 +26,10 @@ class CompaniesHouseService
       Airbrake.notify(e)
       Rails.logger.error "Companies House error: " + e.to_s
       :error
+    rescue SocketError => e
+      Airbrake.notify(e)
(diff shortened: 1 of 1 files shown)


irisfaraway:
@Cruikshanks This is just a really small fix as this was bugging me – not sure it really needs unit tests. But what do you think?

Cruikshanks:
If we can I'd push for a test, again coming back to the specs serving as documentation for how we expect the app to behave.

I tracked this down in one of our gems <https://github.com/DEFRA/ea-address_lookup/blob/master/spec/ea/address_lookup/adapters/address_facade_adapter_spec.rb#L69>

Do you think this would help you spec out what happens in the event the app can't connect?

irisfaraway:
@Cruikshanks Worked like a charm 👍

PR #126, opened 2018-04-13
Ensure each renewals service page has a unique title

irisfaraway:
https://eaflood.atlassian.net/browse/WC-194

We need to go through each page in the renewals service and set a title for each page that is descriptive and unique. Currently all forms have the same title, which is confusing.

WIP


config/locales/en.yml
```diff
@@ -46,8 +49,10 @@ en:
         other: "%{count} addresses found"
 
   # Custom error pages
+  invalid_reg_identifier_heading: Invalid registration number
```
Cruikshanks: Think a typo. Assume it should be `invalid_reg_identifier_title`.

config/locales/forms/register_in_northern_ireland_forms/en.yml
```diff
@@ -1,6 +1,7 @@
 en:
   register_in_northern_ireland_forms:
     new:
+      title: You should register in Northern Ireland
```
Cruikshanks: I'd edge towards dropping the **You should** in the titles of these pages, so go with `Register in Northern Ireland` instead.

What do you think?

config/locales/forms/worldpay_forms/en.yml
```diff
@@ -1,6 +1,7 @@
 en:
   worldpay_forms:
     new:
+      title: Worldpay goes here
```
Cruikshanks: Love it! 😂

config/locales/en.yml
irisfaraway: Oops, yes! Will fix.

PR #127, opened 2018-04-13
Implement version page

irisfaraway:
https://eaflood.atlassian.net/browse/WC-264

This page is a standard feature in our Rails apps that displays the version and commit hash of the currently-running code. This is a diagnostic tool to confirm we are running the correct version, but is also used for external monitoring as a simple, low-impact way of confirming the status of the app.


irisfaraway:
Not entirely sure how to test this, but I'll have a think!

PR #131, opened 2018-04-16
Fall back to title from 'new' action

irisfaraway:
We get custom title attributes from I18n based on the action_name. However, if there isn't one provided for that action, we can fall back to whatever is provided for the 'new' action.

This means that when we hit the 'create' action in a validation error, we can fall back to the title for the 'new' action instead (or provide a custom one if we want).


irisfaraway:
@Cruikshanks This is to fix an issue Tim found during testing – would you mind having a quick peek?

PR #132, opened 2018-04-18
 Checks for can_renew include expires_on

irisfaraway:
https://eaflood.atlassian.net/browse/WC-185

Currently there will be a brief window when a registration is officially EXPIRED but it is still recorded as ACTIVE in the service. This is due to the timing of when the job that marks expired registrations is run.

To avoid any issues with someone renewing before the expiry job runs, we should make sure that a registration cannot be renewed if the current date is equal to or after the expires_on date.

### DIFF
--- a/app/models/concerns/can_change_registration_status.rb
+++ b/app/models/concerns/can_change_registration_status.rb
@@ -42,10 +42,10 @@ module CanChangeRegistrationStatus
       end
 
       event :renew do
-        transitions from: %i[active
-                             expired],
+        transitions from: :active,
                     to: :active,
-                    guard: :close_to_expiry_date?,
+                    guard: %i[close_to_expiry_date?
+                              should_not_be_expired?],
                     after: :extend_expiry_date
       end
     end
@@ -54,6 +54,11 @@ module CanChangeRegistrationStatus
     def close_to_expiry_date?
       expiry_day = registration.expires_on.to_date
       expiry_day < Rails.configuration.renewal_window.months.from_now
+    end
+
+    def should_not_be_expired?
+      expiry_day = registration.expires_on.to_date
+      Date.current < expiry_day
     end
 
     # Transition effects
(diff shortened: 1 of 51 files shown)


irisfaraway:
Tests are passing locally so something must be wrong with the config! Working on it...

app/models/concerns/can_change_registration_status.rb
```diff
@@ -70,4 +77,32 @@ def log_status_change
...
+
+  private
+
+  # expires_on is stored as a Time in UTC and then converted to a Date.
+  # If a user first registered near midnight around the transition between GMT and BST (or the other way round),
+  # there is a risk that the UTC date will not be the same as the UK date.
+  # So we should check for this and compensate to avoid expiring them on the wrong date.
+  def expiry_time_adjusted_for_daylight_savings
+    # If registered in BST and expiring in GMT, add an hour
+    if registered_in_daylight_savings? && !expires_in_daylight_savings?
```
Cruikshanks: It's a minor quibble, but I'd rather see an expressive method name than `if true && !true` expression.

How would you feel about creating a method that expresses what this statement is trying to say e.g. `registered_in_bst_and_expiring_in_gmt?`.

Same goes for the `elsif` below.

spec/models/registration_spec.rb
```diff
@@ -450,6 +450,85 @@
           end
         end
 
+        context "when the registration expiration date is today" do
```
Cruikshanks: welldone

app/models/concerns/can_change_registration_status.rb
irisfaraway: Nope, that totally makes sense! I'll pop that in.

spec/models/registration_spec.rb
Cruikshanks: _Why does Alan never have time to do any work?_ Because he spends what he has trying to find and add just the right gif which expresses how he feels!

irisfaraway: tenor

PR #145, opened 2018-04-26
Add payment summary page

irisfaraway:
https://eaflood.atlassian.net/browse/WC-116

Before they are asked for payment, users should see a summary of the amount they must pay and their options for paying it.

This should include the renewal fee, the cost of any copy cards, and the fee for changing the registration type (if applied).

This does not cover setting up the order or payment objects in the database as this will be dealt with when WorldPay is implemented (WC-117).

Also does not cover routing to the page for paying by bank transfer, as this is a separate story (WC-118) and PR (https://github.com/DEFRA/waste-carriers-renewals/pull/146).


app/forms/payment_summary_form.rb
```diff
@@ -1,17 +1,38 @@
...
-  # TODO: Define accessible attributes, eg attr_accessor :field
+  attr_accessor :temp_payment_method, :type_change, :registration_cards, :registration_card_charge, :total_charge
 
   def initialize(transient_registration)
     super
-    # TODO: Define params to get from transient_registration, eg self.field = @transient_registration.field
+    self.temp_payment_method = @transient_registration.temp_payment_method
+
+    self.type_change = @transient_registration.registration_type_changed?
+    # TODO: Uncomment this once https://github.com/DEFRA/waste-carriers-renewals/pull/139/ is merged in
```
Cruikshanks: It is merged now so we can sort this in this PR.

app/forms/payment_summary_form.rb
```diff
@@ -1,17 +1,38 @@
...
+
+  def determine_total_charge
+    charges = [Rails.configuration.renewal_charge]
+    charges << Rails.configuration.type_change_charge if type_change
+    charges << registration_card_charge
+    charges.sum
+  end
+
+  def determine_registration_card_charge
+    registration_cards * 5
```
Cruikshanks: To keep things consistent I'd go with making this a config value rather than a magic number.

irisfaraway:
@Cruikshanks Good spots! Should be fixed now.

PR #146, opened 2018-04-27
Pay for a renewal by bank transfer

irisfaraway:
https://eaflood.atlassian.net/browse/WC-118

If a user chooses to pay by bank transfer on the Payment Summary page, they should be sent to an additional page which has all the necessary instructions, information and bank details for them to make the payment.

This is instead of going through WorldPay.

It should be made clear that paying by bank transfer means you cannot instantly renew your registration as you have to wait for your renewal to be processed.

Somewhat reliant on https://github.com/DEFRA/waste-carriers-renewals/pull/145

Note also that this doesn't cover progressing to the "renewal application received" page - that's covered by WC-120.


irisfaraway:
@Cruikshanks Think this is almost at review stage, but I'll old off on requesting on until #145 is merged in as I'll need to make some changes after that.

Cruikshanks:
No worries. #145 has a couple of minor request for changes, but ready to approve and move onto this when you say so!

PR #149, opened 2018-04-30
'Application received' page for renewals with a pending conviction check or payment

irisfaraway:
https://eaflood.atlassian.net/browse/WC-120

If a user submits their renewal and pays by bank transfer, or submits and has a conviction check pending, they should see a "Renewal received" page.

This story covers setting up that page and adding routing to it depending on the payment type, or if the user has declared convictions. It does not cover other sorts of conviction checks yet.

### DIFF
--- a/config/locales/forms/renewal_received_forms/en.yml
+++ b/config/locales/forms/renewal_received_forms/en.yml
@@ -4,11 +4,14 @@ en:
       title: Renewal application received
       heading: Application received
       highlight_text: "Your registration number is still"
-      paragraph_1: "Details of how to pay have been emailed to %{email}."
-      subheading_1: "Next step: pay the renewal charge"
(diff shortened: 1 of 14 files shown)


config/locales/forms/renewal_received_forms/en.yml
```diff
@@ -0,0 +1,27 @@
+en:
+  renewal_received_forms:
+    new:
+      title: Renewal application received
+      heading: Application received
+      highlight_text: "Your registration number is still"
+      paragraph_1: "We have sent a confirmation email to to %{email}."
```
Cruikshanks: Typo 😁 "...to to %{email}."

irisfaraway: Oops. Should be fixed now!

PR #154, opened 2018-05-01
associating CBDU23 to another account

timstone:
Just changing as you've now added multiple user seed support


ubuntu-xenial-16.04-cloudimg-console.log
```diff
@@ -0,0 +1,626 @@
+[    00000 Iniializing grp subss ust
```
irisfaraway: I assume this is something to do with our upgraded Vagrant setup 🤔Could you delete it or make sure it's covered by the .gitignore?

timstone: oh good spot, will do.

timstone:
@irisfaraway log file removed, good spot!

PR #162, opened 2018-05-03
Play nice with browser back button

irisfaraway:
https://eaflood.atlassian.net/browse/WC-307

Viewing or submitting forms is currently only allowed if the saved workflow_state matches the form in question.

This should be fine if the user is only navigating by the buttons in the app, but using the browser back button to access earlier, cached forms means the user can be looking at a different form than what the app thinks they should be looking at.

Submitting one of these cached forms will send the uesr back to the form matching the workflow_state, but won't save anything they submit, or warn them about that.

We should modify how the state machine works to make sure users can still navigate with the back button and save their changes.

WIP!


irisfaraway:
Also, CodeClimate is mad because some of the test outlines I've written for the expected future behaviour aren't in use yet, and so those new tests aren't considered to be covered by our test suite ¯\_(ツ)_/¯

irisfaraway:
@Cruikshanks Ready for another review on this I think. Browser back button behaviour should now be adjusted for all pages after the start page through to the 'Check your details' page. Not actually that painful to implement after all!

The next step after this will be improving the validations on the 'Check your details' page, which feels distinct enough that I'd rather get a review on this bit first.

app/helpers/base_forms_helper.rb
```diff
@@ -0,0 +1,40 @@
+module BaseFormsHelper
+  # The standard behaviour for loading a form is to check whether the requested form matches the workflow_state for
+  # the transient registration, and redirect to the saved workflow_state if it doesn't.
+  # But if the workflow state is 'flexible' (listed below), then we skip this check and load the requested form.
+  # This means users can still navigate by using the browser back button and reload forms which don't match the
+  # saved workflow_state. We then update the workflow_state to match their request, rather than the other way around.
+  # These are generally forms after 'renewal_start_form' but before 'declaration_form'.
+  def flexible_states
```
Cruikshanks: Apologies, if I had a bit more time I would have coded up an example so I just have my quick thoughts. My thinking is rather than have a separate list we check the state against, could we not just ask the form/controller itself?

I'm conscious the form and the state are two separate things, even though they may have the same name, and I probably need you to walk me through the flow so I understand it better.

But rather than leave you hanging (as I'll be in probably around lunchtime tomorrow) I'd blurt out my current thoughts so you have a chance to spot where I've missed the point ahead of time 😁!

So I'm looking at `set_up_form` in the forms_controller and wondering could we just make a `form_class.flexible?` call, with `flexible?` being a class method. Or something similar on the controller class? That way when we look at the code for the form/controller we can see there whether we are dealing with flexible form or not, plus we don't need to remember to keep this separate list up to date as and when we make changes to the journey.

Anyway, just my rambles. Other than that nothing else I could think to come back with.

irisfaraway: That makes sense to me and sounds like a much better approach! I'll have a bash at it.

irisfaraway: Had a go at implementing this – what do you think?

app/models/concerns/can_navigate_flexibly.rb
```diff
@@ -0,0 +1,5 @@
+# Used to flag forms as 'flexible' for navigational purposes
```
Cruikshanks: Cracking idea! 🧀 Now I open the class and can see immediately whether its flexible or not. Added bonus; Codeclimate doesn't moan.

Only final request is that I found the comments in the deleted `BaseHelper` useful. Could they be brought into here to give a bit more of an explanation of what we mean by `CanNavigateFlexibly`?

irisfaraway: Good point! I've resurrected the comment.

spec/requests/registration_number_forms_spec.rb
```diff
@@ -4,6 +4,10 @@
   include_examples "GET flexible form", form = "registration_number_form"
 
   describe "POST registration_number_forms_path" do
+    before do
```
Cruikshanks: 🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 
The stubbing makes so much sense when we aren't testing the CH service itself. This is awesome; I hope the effort is rewarded by some faster tests and not having to worry about valid CH numbers.
🦄🦄🦄🦄🦄🦄🦄🦄🦄🦄🦄

irisfaraway:
Had to rebase to deal with the conflict in the VCR cassettes, so unfortunately where reviews happened in relation to the commits has been lost.

irisfaraway:
@Cruikshanks Review round... 4? Annoyingly the rebase has messed up the history but I'm pretty sure it was somewhere in the "Move X validations to custom validator" commits when you last looked at it.

The numbers at the top of this PR are horrifying.

app/models/concerns/can_limit_number_of_people.rb
```diff
@@ -0,0 +1,81 @@
+module CanLimitNumberOfPeople
+  extend ActiveSupport::Concern
+
+  def enough_people?(type = person_type)
+    return false if number_of_existing_people(type) < minimum_people(type)
+    true
+  end
+
+  def can_only_have_one_person?(type = person_type)
```
Cruikshanks: Just out of curiosity, why do we set `type` as `person_type` instead of just using `person_type` in the code?

app/models/concerns/can_limit_number_of_people.rb
```diff
@@ -0,0 +1,81 @@
+module CanLimitNumberOfPeople
```
Cruikshanks: Having looked at this I'm thinking of whether it would be worth (and whether it's possible) to treat this as a base for 2 distinct concerns `CanLimitNumberOfMainPeople` and `CanLimitNumberOfRelevantPeople`.

I think it would remove the need for a lot of code that returns a value based on a type; as distinct concerns they could just refer/use the value relevant to them.

What do you think?

app/forms/check_your_answers_form.rb
```diff
@@ -71,4 +72,20 @@ def displayable_address(address)
      address.postcode,
      address.country].reject
   end
+
+  def should_be_renewed
+    no_invalid_business_type_change? && same_company_no?
+  end
+
+  def no_invalid_business_type_change?
```
Cruikshanks: Just a personal preference (so tell me to sod off) but I believe dropping the `no_` makes the intent clearer. So just `invalid_business_type_change?`. 🤔

app/models/concerns/can_limit_number_of_people.rb
irisfaraway: We want to be able to pass in a specific type if we want – this is mainly because CheckYourAnswers deals with multiple person_types so we need to validate them individually.

spec/support/shared_examples/validate_company_no.rb
```diff
@@ -54,5 +54,27 @@
         expect(form).to_not be_valid
       end
     end
+
+    context "when the business_type doesn't require a company_no" do
+      before do
+        form.business_type = "soleTrader"
+      end
+
+      context "when the company_no is not blank" do
```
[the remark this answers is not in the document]
irisfaraway: Good points.

In these scenarios, what should we do with the lingering company number? Delete it?

app/forms/check_your_answers_form.rb
irisfaraway: I agree, it is awkward! Will try to rephrase this.

app/models/concerns/can_limit_number_of_people.rb
irisfaraway: Fair point. I think I was trying to reduce duplication but it's probably just overcomplicated things! I'll have a bash at splitting them out again.

spec/support/shared_examples/validate_company_no.rb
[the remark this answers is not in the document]
Cruikshanks: Yep, I'd say delete it but I think there could be other situations where we could get data that is no longer appropriate that _should_ be deleted but we won't.

So don't worry if you don't. I think its sufficient to say if the business type is one that requires a number then validate it, else don't worry if its populated.

PR #167, opened 2018-05-23
Pay for renewal using WorldPay

irisfaraway:
https://eaflood.atlassian.net/browse/WC-117

Users should be able to pay for their renewal using WorldPay. We should also handle any payment failures or errors with WorldPay in a graceful way (eg without throwing an error or locking the user out of the process). Once a successful payment has been made using WorldPay, redirect to the 'renewal complete' or 'renewal application received' page (depending on whether a conviction check is pending).


irisfaraway:
@Cruikshanks Here it is! Definitely expecting some suggested changes due to the size and complexity of this – I've been considering shifting more of the logic into the service and out of the controller, so let me know if you agree.

PR #203, opened 2018-06-27
Migrate renewals journey to an engine

irisfaraway:
https://eaflood.atlassian.net/browse/WC-188

The Waste Exemptions (WEX), Flood Risk Activity Exemptions (FRAE), and Project Application and Funding (PAF) services are built such that they are made of 3 projects:

- A front end external web app responsible just for styling and layout
- A back end external web app responsible for styling and layout, plus any additional back office functionality
- A shared 'engine' which contains all the logic and content for the registration/application journey

Waste carriers should follow the same model. To reach this, we need to rework the renewals journey into an engine, which can then be included in the front and back offices.


irisfaraway:
Mounting the engine in https://github.com/DEFRA/waste-carriers-front-office/pull/2

PR #210, opened 2018-07-10
Ensure renewals engine uses names and values as is

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-315

With the upgrade of the services layer of Waste Carriers some of the field names now differ from what the current production system uses, and what the renewals engine also uses.

Plus we've spotted some irregularities in what values are being set. For example the renewals engine is setting the responses to smart answer questions as true or false, but the front end uses "yes" and "no".

So this PR is about getting the 2 in sync enough that a registration created by the frontend can be renewed in the renewals engine.

WIP


Cruikshanks:
Has been dealt with in PR #214

PR #218, opened 2018-07-25
All boolean attributes should be "yes" or "no"

irisfaraway:
https://eaflood.atlassian.net/browse/WC-419

If there are questions where there is a yes or no answer, like smart answers or "Do you have any declared convictions?" we've been saving the answer as a boolean value.

However, the old system saves them as a string - "yes" or "no".

So we have to update the renewals engine to save and validate these answers as strings instead (at least until we can fix our technical debt).


irisfaraway:
Didn't mean to dismiss that review – gonna go ahead and merge this.

PR #225, opened 2018-07-31
WC-336 Adding conviction details paragraph

timstone:
I haven't seen if this is actually displays ok yet so may need your help with that.


app/views/waste_carriers_engine/conviction_details_forms/new.html.erb
```diff
@@ -18,6 +20,7 @@
         <% if @conviction_details_form.errors[:position].any? %>
         <span class="error-message"><%= @conviction_details_form.errors[:position].join(", ") %></span>
         <% end %>
+        <p><%= t(".paragraph_1") %></p>
```
irisfaraway: I think this may have been added to the wrong spot – it's been included in the position fieldset, rather than directly under the h1.

app/views/waste_carriers_engine/conviction_details_forms/new.html.erb
```diff
@@ -7,6 +7,8 @@
 
     <h1 class="heading-large"><%= t(".heading") %></h1>
 
+
+
```
irisfaraway: Can we ditch the extra whitespace?

timstone:
@irisfaraway suggested changes applied.

PR #226, opened 2018-07-31
WC-404 removing negative contractions

timstone:
Changing can't to can not etc


timstone:
@irisfaraway can not changed to cannot

PR #228, opened 2018-07-31
Move users out of engine

irisfaraway:
When we first set up the engine, we included users and related functionality like Devise and CanCanCan. However, it's become clear that the front office and back office users are sufficiently different that we shouldn't be trying to tackle this in the engine and should let the host apps handle it instead.

This PR strips out users, Devise and CanCanCan so they can be reimplemented in the host applications.


irisfaraway:
PRs for reimplementing users in the:

- waste-carriers-front-office: https://github.com/DEFRA/waste-carriers-front-office/pull/24
- waste-carriers-back-office: https://github.com/DEFRA/waste-carriers-back-office/pull/46

PR #230, opened 2018-08-03
Feature/content changes part three

timstone:
Latest content changes update


timstone:
@irisfaraway removed start page content change

PR #231, opened 2018-08-03
Use host when building Worldpay return URLs

irisfaraway:
Previously we used the front-office URL by default. However, this doesn't work when we also need the engine to work in the back office. Instead we set a host environment variable and read from that.


spec/dummy/config/application.rb
```diff
@@ -58,6 +58,7 @@ class Application < Rails::Application
     config.wcrs_frontend_url = ENV["WCRS_FRONTEND_DOMAIN"] || "http://localhost:3000"
     config.wcrs_services_url = ENV["WCRS_SERVICES_DOMAIN"] || "http://localhost:8003"
     config.os_places_service_url = ENV["WCRS_OS_PLACES_DOMAIN"] || "http://localhost:8005"
+    config.host = config.wcrs_renewals_url
```
Cruikshanks: How would this work if the engine was mounted in the back office? Would it still be true that `wcrs_renewals_url == host == success_path`?

PR #232, opened 2018-08-03
Cancelling Worldpay payment returns error message

irisfaraway:
https://eaflood.atlassian.net/browse/WC-394

Rather than showing a more informative message when cancelling a Worldpay payment, we get an 'Invalid response' error. This is caused because we are unable to build a valid MAC for a cancelled order if the user chooses to cancel their payment. In that case, the parameters we get back do not include the paymentAmount, the paymentCurrency or the paymentStatus. Instead we have orderAmount and orderCurrency.

We need to modify the params we receive for a cancelled payment so we can properly validate them, including the MAC.


config/locales/forms/worldpay_forms/en.yml
```diff
@@ -4,8 +4,9 @@ en:
       failure:
         invalid_response: "Invalid WorldPay failure response"
         message:
-          REFUSED: "Your payment has been refused. Please try again or select a different payment method."
+          CANCELLED: "Your payment has been cancelled. Please try again or select a different payment method."
```
Cruikshanks: Not a content bod, but should this read more along the lines

> Your attempt to pay was cancelled.

To my ears it currently sounds like payment was completed, and then subsequently cancelled. What do you 🤔 💭 ?

irisfaraway: Good point. Will change it now!

irisfaraway: @Cruikshanks Aaaand updated.

irisfaraway:
Build is failing because VCR cassettes are out of date. Fixing now: https://github.com/DEFRA/waste-carriers-renewals/pull/235

PR #233, opened 2018-08-03
Feature/content changes part four

timstone:
Latest wave of content changes


config/locales/forms/tier_check_forms/en.yml
```diff
@@ -3,18 +3,9 @@ en:
...
-        paragraph_3: "Upper tier registrations need to be renewed every 3 years for £%{renewal_charge}. Lower tier registrations are free and don’t expire."
-        tier_details_subheading: "What’s the difference between upper and lower tier?"
-        tier_details_paragraph_1: This depends on what waste you deal with and who creates it.
-        tier_details_paragraph_2: "You need an upper tier registration if you do any of the following:"
-        tier_details_list_1:
-          - item: transport other people's waste (for example, a skip company)
-          - item: carry construction or demolition waste
-          - item: buy and/or sell waste or arrange for others to carry it
-        tier_details_paragraph_3: "If your activities are limited to the categories below, you only need a lower tier registration:"
         tier_details_list_2:
```
irisfaraway: I think `tier_details_list_2` is no longer in use, so this can be ripped out as well.

timstone:
@irisfaraway good spot, I've cleaned that reference up.

PR #250, opened 2018-08-22
Create conviction_sign_offs

irisfaraway:
https://eaflood.atlassian.net/browse/WC-376

When a user has declared convictions or a matching conviction is found, the transient_registration should get a new conviction_sign_off object.

This is then used to determine whether a manual conviction check is required or not.


irisfaraway:
Not sure why coverage has dropped – when I run that spec file on its own, I get 100% coverage, but when running the suite it misses a few lines. Will look into it!

In the meantime, I think the rest of it can be reviewed.

irisfaraway:
@Cruikshanks Solved the mystery of the missing test coverage, so this should now be fully reviewable.

app/services/waste_carriers_engine/conviction_data_service.rb
```diff
@@ -0,0 +1,37 @@
+module WasteCarriersEngine
```
Cruikshanks: 👍

PR #253, opened 2018-08-30
Don't say conv check is pending if renewal impossible

irisfaraway:
If a renewal is revoked, it should no longer be possible for it to still need a conviction check.

### DIFF
--- a/app/models/waste_carriers_engine/transient_registration.rb
+++ b/app/models/waste_carriers_engine/transient_registration.rb
@@ -85,6 +85,7 @@ module WasteCarriersEngine
     end
 
     def pending_manual_conviction_check?
+      return false unless metaData.ACTIVE?
       renewal_application_submitted? && conviction_check_required?
     end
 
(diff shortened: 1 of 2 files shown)


app/models/waste_carriers_engine/transient_registration.rb
```diff
@@ -85,6 +85,7 @@ def pending_payment?
     end
 
     def pending_manual_conviction_check?
+      return false unless metaData.ACTIVE?
```
Cruikshanks: Is this right? Sorry, head full of cold and a day covering childcare but isn't it `return false unless metaData.PENDING?`?

If it was ACTIVE wouldn't it have already been approved? Unless `metaData.ACTIVE` comes from the original registration and we do nothing with it, in which case this completely makes sense 😁

irisfaraway: It's the latter! When we copy data to the transient registration, that includes the metaData.status. So if the original reg was active, the transient reg will also be active.

PR #263, opened 2018-09-14
Add deprecated attributes to avoid DB errors

irisfaraway:
https://eaflood.atlassian.net/browse/WC-464

There are a number of attributes which were once in use in the existing system but have since been deprecated. However, some records in the database still have those fields, and attempting to import these records (eg to create a transient_registration) causes an error.

While the ideal solution would be to clean up our data, for now we should allow these attributes so we don't block users from renewing registrations which have them.


Cruikshanks:
`renewalRequested` is the cause of the error in later registrations we are test converting. I couldn't see this listed so thought I'd mention it before you convert.

irisfaraway:
@Cruikshanks Good spot, will add that in!

PR #266, opened 2018-09-17
Send email when renewal application is received and pending

irisfaraway:
https://eaflood.atlassian.net/browse/WC-354

This task covers ensuring an email is sent when a user gets to the end of the renewal process but it is held as pending - this could be because of a pending payment or a conviction check. An email should be sent to confirm we have their details and the journey is complete.


app/views/waste_carriers_engine/renewal_mailer/send_renewal_complete_email.html.erb
```diff
@@ -120,7 +120,7 @@
               </p>
 
               <p style="font-size: 19px; line-height: 1.315789474; margin: 0 0 15px 0;">
-                <%= t(".details.labels.date_activated") %> <%= @registration.metaData.date_activated %>
+                <%= t(".details.labels.date_activated") %> <%= @registration.metaData.date_activated.in_time_zone("London").strftime("%H:%M %d/%m/%Y") %>
```
Cruikshanks: I believe we'll need to strip the time element if we are referring to when the registration becomes active. Basing this on [previous notes](https://github.com/DEFRA/waste-carriers-renewals/wiki/Activation-and-Expiration#activation) we recorded about activation.

If this is in purely because the existing one includes it, I would still strip it out.

PR #272, opened 2018-10-01
Implement separate controller actions for WorldPay responses

irisfaraway:
https://eaflood.atlassian.net/browse/WC-409

Currently we have 2 different actions to deal with WorldPay responses - success or failure. However, there are actually 4 different responses from WorldPay that could lead to 'failure' - the payment could be pending, cancelled or throw an error as well as failing.

We should have separate controller actions to deal with each of these 5 scenarios. This will allow us to respond more appropriately to the different circumstances of a payment not immediately going through.

### DIFF
--- a/app/controllers/waste_carriers_engine/worldpay_forms_controller.rb
+++ b/app/controllers/waste_carriers_engine/worldpay_forms_controller.rb
@@ -47,7 +47,7 @@ module WasteCarriersEngine
 
       order = find_order_by_code(params[:orderKey])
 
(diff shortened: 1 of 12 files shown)


irisfaraway:
At this point we are in code duplication hell – let the refactoring begin.

irisfaraway:
@Cruikshanks Ready for a review. The end result of this should be no different for users, but it does set us up to deal with the individual actions in a different way as part of the next PR.

PR #273, opened 2018-10-01
Generate & attach PDF cert. to confirmation email

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-479

We attach PDF certificates to new registration emails. However, for renewals, we currently ask them to log in.

If the user can't log in and renewed through an AD process instead, they have no way to access the certificate.

This change will generate and attach the updated certificate to the "renewal complete" email so no user needs to log in to access it.


Cruikshanks:
@irisfaraway I shipped a feature!

carlton

app/mailers/waste_carriers_engine/renewal_mailer.rb
```diff
@@ -6,6 +6,9 @@ class RenewalMailer < ActionMailer::Base
     def send_renewal_complete_email(registration)
       @registration = registration
 
+      certificate = generate_pdf_certificate
+      attachments["WasteCarrierRegistrationCertificate-#{registration.regIdentifier}.pdf"] = certificate if certificate
```
irisfaraway: Can we use `reg_identifier` instead of `regIdentifier`?

app/mailers/waste_carriers_engine/renewal_mailer.rb
```diff
@@ -38,5 +41,23 @@ def renewal_received_template
...
+      @presenter = RegistrationPresenter.new(@registration, view_context)
+      pdf_generator = GeneratePdfService.new(
+        render_to_string(
+          pdf: "certificate",
+          template: "waste_carriers_engine/certificates/registration_pdf"
+        )
+      )
+      pdf_generator.pdf
+    rescue StandardError => e
+      Airbrake.notify(e, { registration_no: @registration.regIdentifier }) if defined?(Airbrake)
```
irisfaraway: Ditto about `reg_identifier` vs `regIdentifier`

app/views/waste_carriers_engine/certificates/registration_pdf.html.erb
```diff
@@ -0,0 +1,148 @@
...
+                  <td><%= @presenter.complex_organisation_name %></td>
+                </tr>
+              <% end %>
+              <tr>
+                <td class="table-column-text"><%=t ".registered" %></td>
+                <td><%= @presenter.tier_and_registration_type %></td>
+              </tr>
+              <tr>
+                <td class="table-column-text"><%=t ".registration_number" %></td>
+                <td><%= @presenter.regIdentifier%></td>
```
irisfaraway: Ditto about `reg_identifier` vs `regIdentifier`

spec/mailers/waste_carriers_engine/renewal_mailer_spec.rb
```diff
@@ -34,6 +34,50 @@ module WasteCarriersEngine
...
+        it "has 2 attachments (pdf and logo)" do
+          expect(@mail.attachments.length).to eq(2)
+        end
+
+        it "has an attachment of type pdf" do
+          expect(@pdf_attachment.content_type).to start_with("application/pdf;")
+        end
+
+        it "has a pdf attachment with the right identifier in the filename" do
+          expect(@pdf_attachment.filename).to eq("WasteCarrierRegistrationCertificate-#{@registration.regIdentifier}.pdf")
```
irisfaraway: Ditto about `reg_identifier` vs `regIdentifier`

app/views/waste_carriers_engine/shared/_address_display.html.erb
```diff
@@ -0,0 +1,29 @@
+<%= a.firstName %>  <%= a.lastName %></br>
+<% unless a.houseNumber.blank? %>
```
irisfaraway: A couple thoughts on this:

- These attributes are aliased to use `snake_case`, so would prefer `house_number` to `houseNumber`
- I'd prefer `if foo.present?` instead of `unless foo.blank?`, but I think both might be unnecessary because:
- I recently did some refactoring to reduce duplication in displaying addresses and now we have this `displayable_address` helper method: https://github.com/DEFRA/waste-carriers-renewals/blob/master/app/helpers/waste_carriers_engine/application_helper.rb#L55-L67 Could we use this instead and just display each line of the array returned? I think it might simplify things a bit.

app/views/waste_carriers_engine/certificates/registration_pdf.html.erb
```diff
@@ -0,0 +1,148 @@
...
+            </tbody>
+          </table>
+
+          <h2 class="certificate_title"><%=t ".heading2" %></h2>
+
+          <!-- Renewal details section -->
+          <div class="renewal">
+            <p class="certificate_text"><%= t ".upper_renewal_html", expires_after_pluralized: @presenter.expires_after_pluralized %></p>
+            <% if @presenter.assisted_digital? %>
+              <p class="certificate_text"><%= t ".updatePhone" %></p>
```
irisfaraway: Can we make `updatePhone` snake_case?

app/presenters/waste_carriers_engine/registration_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+    # By complex we mean is there a need to display extra detail in the
+    # document. If its a soletrader we display we display an extra section,
+    # which in the case of partners is a list of their names, and for sole
+    # traders its their business name.
+    def complex_organisation_details?
+      return false unless ["soleTrader", "partnership"].include?(business_type)
+      true
+    end
+
+    def complex_organisation_title
```
irisfaraway: These method names are very similar and a little confusing – what's the difference between a `complex_organisation_title` and a `complex_organisation_name`? Might benefit from some comments or more distinct naming.

app/presenters/waste_carriers_engine/registration_presenter.rb
```diff
@@ -0,0 +1,64 @@
+module WasteCarriersEngine
+  class RegistrationPresenter < BasePresenter
```
irisfaraway: I'm not sure about naming this the RegistrationPresenter – it's only used for certificates at the moment but that name makes it sound like its scope is a lot bigger. If we're intending to start using this logic elsewhere in the engine then I'm fine with it.

spec/presenters/waste_carriers_engine/registration_presenter_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+          presenter = RegistrationPresenter.new(registration, view)
+          expect(presenter.expires_after_pluralized).to eq("3 years")
+        end
+      end
+    end
+
+    describe "#list_main_people" do
+      let(:registration) do
+        reg = create(:registration, :has_required_data)
+        person = build(:key_person, :has_required_data, :main)
```
irisfaraway: I think this could be done entirely with factories instead of building, modifying and then adding a person. For example:

```
first_man = build(:key_person,
                  :has_required_data, 
                  :main, 
                  first_name: "Ryan", 
                  last_name: "Gosling")
second_man = build(:key_person,
                   :has_required_data,
                   :main,
                   first_name: "Corey",
                   second_name: "Stoll)

create(:registration,
       :has_required_data,
       key_people: [first_man, second_man])
```

Gemfile
```diff
@@ -32,6 +32,9 @@ gem "secure_headers", "~> 5.0"
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
 gem "turbolinks"
 
+# Generates a PDF from HTML, in our case, the users certificate
+gem 'wicked_pdf'
```
irisfaraway: Think this should be double quotes.

app/presenters/waste_carriers_engine/base_presenter.rb
```diff
@@ -0,0 +1,27 @@
+module WasteCarriersEngine
+
+  # The `BasePresenter` inherits from SimpleDelegator, and this allows us with
+  # the line `super(model)` in the intializer to say any calls made on the
+  # presenter which are not recognised, pass down to the model passed in. For
+  # example if `RegistrationPresenter` inherits `BasePresentera`, when it is
```
irisfaraway: Typo - `BasePresentera`

app/presenters/waste_carriers_engine/registration_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+      ActionController::Base.helpers.pluralize(
+        Rails.configuration.expires_after,
+        I18n.t("#{LOCALES_KEY}.year")
+      )
+    end
+
+    def list_main_people
+      list = key_people
+        .select { |person| person.person_type == 'KEY' }
+        .map    { |person| format('%s %s', person.first_name, person.last_name) }
```
irisfaraway: Should this be double quotes around `%s %s`?

app/views/waste_carriers_engine/certificates/registration_pdf.html.erb
```diff
@@ -0,0 +1,148 @@
...
+              <tr>
+                <th><%=t ".authority" %></th>
+                <td></td>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td class="table-column-text"><%=t ".name" %></td>
+                  <td>
+                    <%= wicked_pdf_image_tag("environment_agency_logo.png", {:alt => "Environment Agency Logo", :border => "0", :height => "39", :width => "129"}) %>
```
irisfaraway: Guessing this came from the existing template, but can we use the new hash syntax here?

app/views/waste_carriers_engine/certificates/registration_pdf.html.erb
```diff
@@ -0,0 +1,148 @@
...
+              </tr>
+              <tr>
+                <td class="table-column-text"><%=t ".business_address" %></td>
+                <td>
+                  <%= render "waste_carriers_engine/shared/address_display.html.erb", r: @presenter, a: @presenter.registered_address  %>
+                </td>
+              </tr>
+              <tr>
+                <td class="table-column-text"><%=t ".carrier_telephone_number" %></td>
+                <td><%= @presenter.phoneNumber%></td>
```
irisfaraway: Think we can use `phone_number` as it's aliased.

spec/presenters/waste_carriers_engine/registration_presenter_spec.rb
irisfaraway: Just noticed we do the same thing to test `complex_organisation_name` - might be an opportunity to reduce some duplication.

Cruikshanks:
Thx @irisfaraway . Think I did try some of the snake case versions of attributes and it failed, but that might just be the fog of bashing through another issue.

And defo my bad on some things; was a copy & paste job from frontend so I've let some of its taint creep in.

Cruikshanks:
@irisfaraway We discussed changing the email content however I double checked what we do for new registrations, and the content is exactly the same. So for now I think it's sufficient to leave it as is.

N.B. proof its now working (for me at least!)

screen shot 2018-10-16 at 14 29 03

PR #275, opened 2018-10-02
Log WorldPay responses

irisfaraway:
Log whether the response is valid or not, the response type from WorldPay (eg 'success', 'cancel') and the params which produced this outcome.

### DIFF
--- a/app/controllers/waste_carriers_engine/worldpay_forms_controller.rb
+++ b/app/controllers/waste_carriers_engine/worldpay_forms_controller.rb
@@ -15,17 +15,11 @@ module WasteCarriersEngine
     def create; end
 
     def success
-      return unless set_up_valid_transient_registration?(params[:reg_identifier])
+      respond_to_acceptable_payment(:success)
+    end
 
-      if worldpay_service(params).valid_success?
-        log_and_send_worldpay_response(true, :success)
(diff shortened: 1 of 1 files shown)


irisfaraway:
@Cruikshanks Think we may have to be picky about which ones we log – for example, I don't know if it would be useful to log `cancel` responses. But we could certainly do it all as a temporary measure until we have a better idea of the issue.

Cruikshanks:
@irisfaraway That was my thinking. Once we have a better handle on our Worldpay fails, and whether a new solution is needed, we can re-evaluate what logging we do need around this.

And on being picky, I still think I'd like to get notifications on cancelled ones for now. It might be that users are cancelling but something about how it happens means it doesn't cancel cleanly. If i can link reported registrations with errbit notifications (hint-hint getting the reg number in the notification would be super helpful 🙏 ) I can then more easily determine where the issues lie.

irisfaraway:
@Cruikshanks Don't mind if you review this or #274 first, but I'll need to refactor whichever one is the second PR so probably worth holding off on the second review until that's done. Pick your favourite!

irisfaraway:
@Cruikshanks Had to do some further refactoring to reduce duplication as raised by CodeClimate – I grumbled, but it was right.

PR #279, opened 2018-10-09
Fix validation on copy card page

irisfaraway:
https://eaflood.atlassian.net/browse/WC-498

Our validation for the `temp_cards` field on the copy card form assumed that leaving it blank would sub in a zero, because the field is classed as an Integer in the database.

This is not the case - if the field already existed, the value is replaced with null, and if it doesn't exist then it's not created.

This means that when the order_items are created, checks against the value of the field fail and we get an error.

### DIFF
--- a/app/forms/waste_carriers_engine/cards_form.rb
+++ b/app/forms/waste_carriers_engine/cards_form.rb
@@ -9,6 +9,7 @@ module WasteCarriersEngine
 
     def submit(params)
       # Assign the params for validation and pass them to the BaseForm method for updating
+      # If temp_cards is blank, sub in 0 so it passes validation
       self.temp_cards = if params[:temp_cards].present?
                           params[:temp_cards]
                         else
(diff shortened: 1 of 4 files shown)


app/forms/waste_carriers_engine/cards_form.rb
```diff
@@ -9,15 +9,18 @@ def initialize(transient_registration)
...
+                        else
+                          0
+                        end
       attributes = { temp_cards: temp_cards }
 
       super(attributes, params[:reg_identifier])
     end
 
     # Must be a positive integer or 0
     # Leaving it blank is also valid - this will automatically sub in a 0 since the field is marked as an Integer
```
Cruikshanks: I'd suggest updating the comment as well as we now know this is not true.

irisfaraway: Oops, good point! Will fix

PR #280, opened 2018-10-09
Duplicate house number in addresses

irisfaraway:
https://eaflood.atlassian.net/browse/WC-500

Addresses created in the renewals journey can display as having a duplicate house number when they are used by waste-carriers-frontend, for example:

123
123 Waste Street
Wastetown
AB1 2CD

This is because the renewals engine assigns the values of address_lines in a different way. For consistency's sake, we should change this to match how waste-carriers-frontend does it.

---

🔥 WIP 🔥

### DIFF
--- a/app/models/waste_carriers_engine/address.rb
+++ b/app/models/waste_carriers_engine/address.rb
@@ -73,16 +73,14 @@ module WasteCarriersEngine
     end
 
     def assign_address_lines(data)
-      lines = []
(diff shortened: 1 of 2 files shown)


irisfaraway:
Replacing with https://github.com/DEFRA/waste-carriers-renewals/pull/281 as it turns out waste-carriers-frontend behaves a little differently.

PR #288, opened 2018-10-22
Fix broken HighVoltage routing

irisfaraway:
https://eaflood.atlassian.net/browse/WC-516

We were trying to put HighVoltage pages at the root, but this resulted in them not being accessible when mounted in an application. Allowing them to sit at "/pages" as normal means they can be accessed from "/fo/pages" and "/bo/pages", which fixes the issue.


irisfaraway:
CI is knackered due to https://status.github.com/messages - will try to kick the checks off again once everything is working.

PR #290, opened 2018-10-27
Add a grace window for expired regs to renew

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-512

Currently when the registration expires there is no mechanism to still start the renewal. Our NCCC team have made a strong case that this is problematic for a number of reasons, and as such they are finding they are having to implement a workaround which involves

- creating a new registrations for users whose reg has expired (customer not happy they have lost their reg. number)
- refunding the user for the difference between a new reg and a renewed one (NCCC and finance not happy because of the extra time and effort this takes)

The reasons given for still permitting a user to renew are

- Current content around the expiry date in both the service and the EPR generally just says "Expiry date". So for example "Expiry date: October 15 2018". However NCCC report that its a common complaint to them that users read this as it expires at October 15 midnight, and as such is still valid to renew on the 15th.
- Currently with no self serve mechanism to transfer or change the account email of a registration, users who have an expiry date on a weekend, may attempt to renew on the Friday, find they are locked out, and have to wait to the next business day to report the issue, by which time they have expired.

So this change is about still allowing registrations which have expired to renew, as long as that expiry falls within a predefined "grace window".


Cruikshanks:
@irisfaraway Fixed my failing tests, and they all needed fixing in the same way. I'm happy I've kept in keeping with what the test was trying to do, but got a niggle in the back of my mind that the way of solved the problem might not be the best. Let me know if you think there is a better way.

app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
```diff
@@ -100,9 +99,24 @@ def should_not_be_expired?
       expiry_day = expiry_time_adjusted_for_daylight_savings.to_date
       # We store dates and times in UTC, but want to use the current date in the UK, not necessarily UTC
       current_day = Time.now.in_time_zone("London").to_date
+
+      return true if in_expiry_grace_window?(expiry_day, current_day)
```
irisfaraway: I think this addition changes the meaning of the method and makes the name inaccurate. This method was initially intended to check whether the registration's status should be 'EXPIRED' and would be changed as soon as the scheduled job hit it, thus the concept of "should be expired".

However, now it could still return `true` if it should be expired _but_ happens to be in the grace window. This feels inaccurate to me. I think we are conflating the concepts of a registration being expired and not being renewable, but they're not the same thing.

So I think it would be better to call `in_expiry_grace_window?` from `renewal_allowed?` instead of putting it in this method.

This may mean a bit of additional refactoring to reduce duplication of local `expiry_day` variables (which I notice are inconsistently calculated anyway - a pre-existing problem though!) but I'm OK with that. What do you think?

app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
```diff
@@ -100,9 +99,24 @@ def should_not_be_expired?
...
     end
 
+    # Its important to note that a registration is expired on its expires_on date.
+    # For example if the expires_on date is Oct 1, then the registration was
+    # ACTIVE Sept 30, and EXPIRED Oct 1. If the grace window is 3 days, just
+    # adding 3 days to that date would give the impression the grace window lasts
+    # till Oct 4 (i.e. 1 + 3) when in fact we need to include the 1st as one of
+    # our grace window days.
+    def in_expiry_grace_window?(expiry_date, current_day)
+      return false if current_day > ((expiry_date + Rails.configuration.grace_window) - 1.day)
```
irisfaraway: Think this line would be clearer if we calculated `((expiry_date + Rails.configuration.grace_window) - 1.day)` on a separate line and then compared the result to `current_day`, similar to the frontend: https://github.com/DEFRA/waste-carriers-frontend/pull/172/files#diff-85bbbdcab8cf9243156faf3f9a3873b8R44

config/application.rb
```diff
@@ -66,6 +66,11 @@ class Application < Rails::Application
     # Times
     config.renewal_window = ENV["WCRS_REGISTRATION_RENEWAL_WINDOW"].to_i
     config.expires_after = ENV["WCRS_REGISTRATION_EXPIRES_AFTER"].to_i
+    # Expired Upper tier registrations can still be renewed if the current date
+    # is within a given 'grace window' after the registration expired. So if the
+    # window is 3 days, the current date is October 12, and the reg. expired Oct
+    # 10 then the reg. is within the window and can still be renewed.
+    config.grace_window = ENV['WCRS_REGISTRATION_GRACE_WINDOW'].to_i
```
irisfaraway: Double quotes!

spec/requests/waste_carriers_engine/renewal_start_forms_spec.rb
```diff
@@ -35,7 +35,11 @@ module WasteCarriersEngine
               end
 
               context "when the registration cannot be renewed" do
-                before(:each) { registration.metaData.expire! }
+                before(:each) do
+                  registration.metaData.expire!
```
irisfaraway: If the registration is not renewable once it's past the grace period (regardless of status), wonder if we could make it unrenewable with a single line that also persists the change: `registration.update_attributes(expires_on: Date.today + Rails.configuration.grace_window)`

PR #296, opened 2018-11-06
Fix conviction check to handle grace window

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-512

During testing of the grace window it was spotted that if a renewal was expired, and it matched during conviction checks, then it could not be continued because the button to access convictions approvals did not display.

This was because `TransientRegistration.pending_manual_conviction_check?` had a guard clause that checked the status was `ACTIVE`. This will no longer be the case for registrations which are EXPIRED but in the grace window, hence we change it to use `may_renew?`.

We also learnt doing this that `may_renew?` will only work if called on a `metaData` object that is linked to a registration, because of the call to `registration.identifier` in `CanChangeRegistrationStatus.renewal_application_submitted?`. Hence we had to tweak the tests and code to account for this.

Finally a test in the transient registration spec flagged that it was possible to for `may_renew?` to return true even if the registration in question was REVOKED. Hence a new guard in `CanChangeRegistrationStatus.renewal_allowed?` to check that the status of the registration is one that can be renewed.

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
+++ b/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
@@ -78,17 +78,13 @@ module WasteCarriersEngine
     # Guards
     def renewal_allowed?
       return true if renewal_application_submitted?
-      return true if renew_expired_registration?
-
-      close_to_expiry_date? && should_not_be_expired?
-    end
-
-    def renew_expired_registration?
       return true if in_expiry_grace_window?
 
       # The only time an expired registration can be renewed is if the application has previously been submitted,
       # or it is withion the grace window - otherwise expiry is an automatic no
       return false if EXPIRED?
+
+      close_to_expiry_date? && should_not_be_expired?
     end
 
     def renewal_application_submitted?
--- a/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
+++ b/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
@@ -75,6 +75,7 @@ module WasteCarriersEngine
 
     # Guards
     def renewal_allowed?
+      return false unless %w[ACTIVE EXPIRED].include?(status)
       return true if renewal_application_submitted?
       return true if in_expiry_grace_window?
 
--- a/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
+++ b/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
@@ -77,13 +77,17 @@ module WasteCarriersEngine
     def renewal_allowed?
       return false unless %w[ACTIVE EXPIRED].include?(status)
       return true if renewal_application_submitted?
+      return true if renew_expired_registration?
+
+      close_to_expiry_date? && should_not_be_expired?
+    end
+
+    def renew_expired_registration?
       return true if in_expiry_grace_window?
 
       # The only time an expired registration can be renewed is if the application has previously been submitted,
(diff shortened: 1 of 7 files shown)


app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
```diff
@@ -75,6 +75,7 @@ def log_status_change
 
     # Guards
     def renewal_allowed?
+      return false unless %w[ACTIVE EXPIRED].include?(status)
```
irisfaraway: This line shouldn't be necessary as it's already covered by line 45 – can we try removing it and seeing if it still works?

spec/models/waste_carriers_engine/registration_spec.rb
```diff
@@ -520,9 +530,9 @@ module WasteCarriersEngine
               expect(registration.metaData).to allow_event :renew
             end
 
-            it "expires when it reaches the expiry date in the UK" do
-              # Skip ahead to the start of the day a reg should expire
-              Timecop.freeze(Time.find_zone("London").local(2018, 10, 27, 0, 1))
+            it "expires when it reaches the expiry date plus 'grace window' in the UK" do
+              # Skip ahead to the start of the day a reg should expire, plus the
```
irisfaraway: Think this comment is missing an ending!

irisfaraway: Also, it expires when it hits the expiry date – it can still be renewed for another few days, but that doesn't mean it's not expired. So I'd prefer "it cannot be renewed" to "it expires". Same goes for the test on line 496.

spec/models/waste_carriers_engine/registration_spec.rb
```diff
@@ -452,14 +452,22 @@ module WasteCarriersEngine
...
           context "when the registration expiration date is today" do
-            let(:registration) { build(:registration, :is_active, expires_on: Date.today) }
+            context "and the 'grace' window is 3 days" do
+              before { allow(Rails.configuration).to receive(:grace_window).and_return(3) }
 
-            it "cannot be renewed" do
-              expect(registration.metaData).to_not allow_event :renew
+              let(:registration) { build(:registration, :is_active, expires_on: Date.today) }
+
+              it "cannot be renewed in 3 days time" do
```
irisfaraway: Maybe this is personal preference, but I think this might be clearer in a separate context – I'd prefer "when the registration expiration date was three days ago, it cannot be renewed today" to "when the registration expiration date is today, it cannot be renewed in 3 days' time".

app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
```diff
@@ -77,13 +77,17 @@ def log_status_change
...
+
+      close_to_expiry_date? && should_not_be_expired?
+    end
+
+    def renew_expired_registration?
       return true if in_expiry_grace_window?
 
       # The only time an expired registration can be renewed is if the application has previously been submitted,
       # or it is withion the grace window - otherwise expiry is an automatic no
       return false if EXPIRED?
```
irisfaraway: Now that this isn't followed by further checks, there's nothing to return `true` if it's not `EXPIRED?` – I think we would get `nil` instead. Which makes me wonder if the tests are covering this.

Cruikshanks:
N.B. removing the unnecessary check in `renewal_allowed?` has also meant it can go back to how we previously had it with CodeClimate kicking off.

PR #300, opened 2018-11-13
Allow expired renewal to complete post declaration

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-526

We know now that we have renewals linked to registrations that have expired during the payment phase.

In a lot of these cases its because Worldpay has failed, and we have asked NCCC to use the workaround of sending the renewal back to the payment summary page and then completing it as a BAC's payment. The problem we are now encountering in greater numbers is renewals are expiring before the missed payment can be spotted and the workaround enacted.

We have therefore agreed that once a renewal has had the declaration accepted, it can still be renewed even if it then expires. This change covers the amendments necessary to support this.

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
+++ b/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
@@ -82,7 +82,6 @@ module WasteCarriersEngine
 
     # Guards
     def renewal_allowed?
-      return true if renewal_application_submitted?
       return true if renewal_declaration_confirmed?
 
       # The only time an expired registration can be renewed is if the application has previously been submitted,
--- a/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
+++ b/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
@@ -84,8 +84,10 @@ module WasteCarriersEngine
     def renewal_allowed?
       return true if renewal_declaration_confirmed?
 
-      # The only time an expired registration can be renewed is if the application has previously been submitted,
-      # or it is withion the grace window - otherwise expiry is an automatic no
+      # The only time an expired registration can be renewed is if the
+      # application
+      # - has a confirmed declaration i.e. user reached the copy cards page
+      # - it is withion the grace window
       check_service = ExpiryCheckService.new(registration)
       return true if check_service.in_expiry_grace_window?
       return false if check_service.expired?
--- a/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
+++ b/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
@@ -83,6 +83,7 @@ module WasteCarriersEngine
     # Guards
     def renewal_allowed?
       return true if renewal_application_submitted?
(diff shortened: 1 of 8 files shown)


Cruikshanks:
@irisfaraway I started thinking I was going to want to first ensure the reg was expired before then checking for the declaration, hence the introduction of `ExpiryCheckService` and the refactoring of some methods off the concern into it.

However once I'd finished that realised that wasn't actually the case, so could have got away with just sticking the check.

I prefer having the ExpiryCheckService and all the GMT/BST methods as part of that rather than the concern they were on. However if you asked me to split this PR to separate the refactoring and the actual feature, I'd be listening.

Other than that have tested this manually via the UI and seems to behave as expected. Not made any front-office/back-office PR's as they just need to bump their versions of waste_carriers_engine to these when we're done.

app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
```diff
@@ -78,82 +83,34 @@ def log_status_change
     # Guards
     def renewal_allowed?
       return true if renewal_application_submitted?
```
irisfaraway: I think we can skip the `renewal_application_submitted?` check now, as nothing should make it to `renewal_received` without the declaration being accepted.

However we do use that method for other things, so it should continue to exist - just not be used here.

app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
```diff
@@ -78,82 +83,34 @@ def log_status_change
...
-      return true if in_expiry_grace_window?
+      return true if renewal_declaration_confirmed?
 
       # The only time an expired registration can be renewed is if the application has previously been submitted,
```
irisfaraway: Think this comment needs updating, as it can now be renewed once the declaration is accepted, not just if it's been submitted.

app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
```diff
@@ -6,6 +6,11 @@ module CanChangeRegistrationStatus
     include Mongoid::Document
     include CanCalculateRenewalDates
 
+    # The pattern for denoting events here is common when using AASM, however it does
+    # mean we get flagged by rubocop. We don't want to change the rule at a global
+    # level, but the ecessive block length is acceptable here hence we're happy to add
```
irisfaraway: Small typo here in `excessive`

app/services/waste_carriers_engine/expiry_check_service.rb
```diff
@@ -0,0 +1,106 @@
...
+    def initialize(registration)
+      raise "ExpiryCheckService expects a registration" if registration.nil?
+
+      @registration_date = registration.metaData.date_registered
+      @expires_on = registration.expires_on
+      @expiry_date = corrected_expires_on
+    end
+
+    # For more details about the renewal window check out
+    # https://github.com/DEFRA/waste-carriers-renewals/wiki/Renewal-window
```
irisfaraway: Should we link to dst-guides instead now?

PR #302, opened 2018-11-16
Improve dashboard search performance

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-531

In production our users have reported that when attempting to apply a filter in the dashboard, it becomes unresponsive and eventually leads to a "service unavailable" page.

Initial investigation has shown that the problem is each time a user searches the app is pulling back all transient registrations and then iterating through them,  checking for the matching criteria.

One watching the logs this actually results in the ORM Mongoid querying for each transient registration. Hence its not putting a massive load on MongoDB, but it is now taking so long that the browsers is timing out leading to the service unavailable page.

This change is about improving the performance of the search in the dashboard, such that it no longer times out, and ideally is a scalable solution i.e. we want be back here again in another 6 months.

### DIFF
--- a/app/models/waste_carriers_engine/transient_registration.rb
+++ b/app/models/waste_carriers_engine/transient_registration.rb
@@ -30,6 +30,7 @@ module WasteCarriersEngine
              "addresses.postcode": /#{term}/i)
     }
     scope :submitted, -> { where(:workflow_state.in => %w[renewal_complete_form renewal_received_form]) }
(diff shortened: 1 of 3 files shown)


Cruikshanks:
I think I said in a commit, I would have liked to to have resolved rubocop's issue with the lambda over multiple lines, but the moment I change it Mongoid errors.

I'd also have liked to move the scopes to a concern, to help reduce the size of the class here, but Mongoid didn't seem to like that either. So this works, Code Climate isn't shouting, hence over to you @irisfaraway for review.

PR #307, opened 2018-11-23
Refactor logic around can renew checks

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-533

WC-533 highlighted that we were doing checks for whether something can be renewed in 2 different places, in 2 different ways. We agreed to move the checks to one place, but additional review has highlighted that we're using our current checks for mixed purposes, and as such are not being very efficient.

Currently `app/models/concerns/waste_carriers_engine/can_change_registration_status.rb` has the `renewal_allowed?` which is hooked up to the AASM event `renew`. Hence within the code you can call `metaData.may_renew?`.

You call the renew event when you wish to transition a registration to be 'ACTIVE`, and to extend the existing expiry date and update the ac tivation timestamps. In essence you call this when you are completing a renewal.

However these changes are only part of what happens to a registration when you complete a renewal. Hence we have the `app/services/waste_carriers_engine/renewal_completion_service.rb`. This highlights some of the inefficiencies. For example

- When you are ready to complete a renewal you instantiate a `RenewalCompletionService` passing in the transient registration. In the initializer it also hits the db to pull back the matching registration
- You then call `complete_renewal` which ends up calling `metaData.may_renew?`
- This doesn't have a reference to the transient registration, so makes another hit on the database for that
- It then carries out its checks and returns true or false
- If true we're back in `complete_renewal` making a series of changes including a call to `update_registration`
- In `update_registration` we end up again calling `metaData.renew` (in the guise of requesting a transition not just querying if we can) which in turn repeats the same checks we've already performed, and hits the db for a 3rd time to get the transient registration
- Also in the current form, `RenewalCompletionService` isn't checking if there are any pending payment or conviction checks, as these were being done by the back office.

So this is an exploratory PR, in such that it will look to do 2 things

- split out the checks into the correct contexts; can I start a renewal, can I continue a renewal, and can I complete a renewal
- make the checks more efficient such that we reduce the number of calls to the db for data we already have

It is likely this will be split out into smaller PR's, but for its grouping the changes whilst we understand what changes need to be made.

---

🎉 WIP 🎉


Cruikshanks:
Apologies for the size @irisfaraway , and I'm sure there are commits in here that were done elsewhere, though no amount of rebasing will shift them.

However let me know if you have any questions or issues, and if you'd like a run through of the changes first.

Also I've tested with the front office and it was fine with just updating the reference to this particular branch. Waste carriers back office needed specific changes which you can find in [PR 216](https://github.com/DEFRA/waste-carriers-back-office/pull/216) if you want to give it a run through there.

spec/factories/transient_registration.rb
```diff
@@ -87,8 +93,25 @@
 
     trait :has_revoked_registration do
       # Create a new registration when initializing so we can copy its data
-      initialize_with { new(reg_identifier: create(:registration, :has_required_data,
-      metaData: build(:metaData, revoked_reason: "foo")).reg_identifier) }
+      initialize_with do
```
irisfaraway: I think we can make more use of some of the existing registration factory traits for these new transient_registration traits (like `:is_revoked`).

I also wonder if it'd be better to have new registration traits like `:expired_one_month_ago` to do the work rather than specifying details in these `initialize_with` blocks, as I think this would be more consistent with other traits in this factory.

.travis.yml
```diff
@@ -3,6 +3,7 @@ env:
     - CC_TEST_REPORTER_ID=38776e3c20a07980a0cc3c66c7519e9b42542a8f07d5bcc9d7498d02cc00a612
     - WCRS_REGISTRATION_EXPIRES_AFTER=3
     - WCRS_REGISTRATION_RENEWAL_WINDOW=3
+    - WCRS_REGISTRATION_GRACE_WINDOW=5
```
irisfaraway: I agree with you that we've been inconsistent about stubbing env vars vs putting them in the Travis config and need  to decide on one. I would prefer stubbing as it means the tests should still run as expected in any environment – for example, if we changed our CI or ran the suite on a different machine without setting up the env vars. Maybe something to log as an issue for standardising down the line?

app/models/waste_carriers_engine/transient_registration.rb
```diff
@@ -114,6 +113,34 @@ def pending_manual_conviction_check?
       renewal_application_submitted? && conviction_check_required?
     end
 
+    def can_be_renewed?
```
irisfaraway: I think we maybe need a little more differentiation between `can_be_renewed?` and `ready_to_complete?`. Maybe some more specific naming would help?

My understanding is that `can_be_renewed?` should return true when it's OK for the user to proceed with the renewal process, as opposed to the transient_registration actually being renewable at that time. For that matter, it's not the transient_registration that renews, but the registration.

Whereas `ready_to_complete?` should only return true when the transient_registration is in such a state that renewal can now actually occur.

Is my understanding correct here?

app/services/waste_carriers_engine/renewal_completion_service.rb
```diff
@@ -92,17 +116,17 @@ def merge_finance_details
...
+    def initialise_finance_details(registration)
       registration.finance_details = FinanceDetails.new unless registration.finance_details.present?
-      set_up_orders(registration)
-      set_up_payments(registration)
+      intialise_orders(registration)
+      initialise_payments(registration)
     end
 
-    def set_up_orders(registration)
+    def intialise_orders(registration)
```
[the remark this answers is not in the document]
Cruikshanks: Good spot and agree on using the American version of the spelling. I also think the way Americans say route is correct, so I'm open to controversy 😁

spec/factories/finance_details.rb
```diff
@@ -14,5 +16,22 @@
       orders { [build(:order, :has_required_data)] }
       payments { [build(:payment)] }
     end
+
+    trait :has_paid_order_and_payment do
+      orders { [build(:order, :has_required_data)] }
+      payments do
+        [
+          build(:payment, :bank_transfer, amount: 105_00),
```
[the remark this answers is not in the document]
Cruikshanks: That's just a combination of being useless at maths and just trying to make rubocop happy 😳

spec/factories/transient_registration.rb
Cruikshanks: Good point, though I think we have a general mix of using before/let blocks, or using Factorybot throughout the tests.

Would you be happy that to get this into production (as we have users with stuck registrations waiting on this)

- Where we have an existing trait I update the tests to use it
- Where we don't we leave as is and add an issue to review our factories and the traits required

.travis.yml
Cruikshanks: I would prefer stubbing as well. I'll add an issue, but as I found this by accident rather than design I'd want to stick with leaving the Travis config, and log an issue to remove them all from travis and deal with the fall out then.

app/models/waste_carriers_engine/transient_registration.rb
Cruikshanks: Yes. Where I got a bit stumped was that we referred to the original state machine `:renew` event on the metaData object for every transition because its called in the base `FormsController` (`metaData.may_renew?`). This includes when you do the GET request on the Renewal start page.

So ideally in the GET request for that page I'd want to call `can_be_renewed?` on the registration object. Then for every subsequent page i.e. when we have a saved transient_registration object, I'd want to call something like `can_progress_renewal?` on the transient_registration object.

Moving where we do the checks is one thing, but I felt like a refactor like that was a step to far for the scope of this PR. And in light of your main comment might have been wasted as well.

So we didn't really differentiate the checks before (we just called `may_renew?` in both cases) so I'm happy to leave as is. I think your main point about different state machines and transitions thought for the 2 registrations is where we should focus our efforts.

spec/factories/transient_registration.rb
irisfaraway: Sounds good 👍

.travis.yml
irisfaraway: Yep, agreed!

spec/factories/transient_registration.rb
Cruikshanks: Have updated the code and also created issue #322

.travis.yml
Cruikshanks: Issue #323 has been created for this

Cruikshanks:
Have agreed with @irisfaraway that the main concern she raises is correct but that we are going to ship this in its current state.

We need to conduct a proper review of the existing state machine we are using and have created an issue for that (see #324 ). But we also have to acknowledge that we have renewals that are stuck with users waiting for them to be completed, and we need the changes in this refactor to resolve them.

Cruikshanks:
Have agreed with @irisfaraway that the main concern she raises is correct but that we are going to ship this in its current state.

We need to conduct a proper review of the existing state machine we are using and have created an issue for that (see #324 ). But we also have to acknowledge that we have renewals that are stuck with users waiting for them to be completed, and we need the changes in this refactor to resolve them.

PR #311, opened 2018-11-28
Remove may_renew? from pending_manual_conviction_check?

Cruikshanks:
A transient registration is pending a conviction check if

- its submitted
- the user declared a conviction, or the automated checker matched the organisation or an individual

Previously it was concerning itself with whether it could also be renewed. This is most likely due to the fact prior to the introduction of named mongoid scopes for searching, this would be used to filter out those that needed an approval to continue.

This change makes the decision that whether a registration can be renewed is irrelevant to the question are there outstanding conviction checks for this transient registration.

It also means one less hit on the db 😁

### DIFF
--- a/app/models/waste_carriers_engine/transient_registration.rb
+++ b/app/models/waste_carriers_engine/transient_registration.rb
@@ -122,9 +122,6 @@ module WasteCarriersEngine
     end
 
     def pending_manual_conviction_check?
-      registration = Registration.where(reg_identifier: reg_identifier).first
-      return false unless registration.metaData.may_renew?
-
       renewal_application_submitted? && conviction_check_required?
     end
 


irisfaraway:
@Cruikshanks I was trying to work out why this now-removed line was here, and I think it may have been so a renewal which had been revoked during the conviction check would not display messages in the back office about needing a convictions check.

See https://github.com/DEFRA/waste-carriers-back-office/blob/9fc05cda60e853f99787a53b22e0c103e864a14a/app/views/dashboards/index.html.erb#L93 and https://github.com/DEFRA/waste-carriers-back-office/blob/63447516d87aaa44089920ba4c266914dcd87cea/app/views/transient_registrations/show.html.erb#L23

I think this is the right change to make so approving anyway, but we probably have to update the logic in the back office to reflect this.

PR #314, opened 2018-11-28
Move methods from TransientRegistration to concern

Cruikshanks:
Whilst working with the `TransientRegistration` spotted that there are a couple of methods that could just as easily sit in the shared concern between it and `Registration`.

- `charity?()`
- `company_no_required?()`

So this change covers moving them over to the concern.


Cruikshanks:
On hold to fix the codeclimate issues because I'm waiting for PR #313 to be merged, which brings in `can_have_registration_attributes.rb` and holds the tests for things on the concern.

PR #319, opened 2018-11-30
Fix frozen string literal issue in a spec file

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-535

Found this issue in the address_spec and have fixed it.


spec/models/waste_carriers_engine/address_spec.rb
```diff
@@ -1,4 +1,6 @@
-require "rails_helper"
+# frozen_string_literal: true
+
+require 'rails_helper'
```
irisfaraway: These should be double quotes.

PR #320, opened 2018-11-30
Add missing # frozen_string_literal: true to spec

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-535

As one off task this change ensures all files within the spec folder have the missing `# frozen_string_literal: true` added.

This automatically means a large number of the files in this folder no longer fall foul of rubocop.


Cruikshanks:
This breaks lots of things so not going to bother right now!

PR #345, opened 2019-01-03
Update CanStripWhitespace to work with frozen strings

irisfaraway:
In Ruby 3, all string literals are immutable. This isn't true in Ruby 2, but to prepare for an eventual upgrade, we are adding the comment `# frozen_string_literal: true` to the start of all our files. Rubocop enforces this.

However, while adding this comment, we discovered the `string.strip!` command in our CanStripWhitespace concern violates this, as it attempts to modify a frozen string.

This PR refactors the concern to work with frozen strings.


irisfaraway:
This doesn't have any immediate effect on its own, but tried using this fix on another branch (https://github.com/DEFRA/waste-carriers-renewals/pull/344) and it resolved the issue with adding the frozen string comment.

PR #351, opened 2019-01-04
Fix style issues in models

irisfaraway:
This PR fixes up style issues in our models so we can enable Rubocop CI.


app/models/waste_carriers_engine/transient_registration.rb
```diff
@@ -24,7 +25,7 @@ class TransientRegistration
     field :temp_payment_method, type: String
     field :temp_tier_check, type: String # 'yes' or 'no' - should refactor to boolean
 
-    scope :search_term, ->(term) {
+    scope :search_term, lambda { |term|
```
Cruikshanks: I swear I did this when I implemented the scopes, but it just kept erroring! 😖😕

PR #402, opened 2019-04-09
Limit extra cards order to 999

cintamani:
https://eaflood.atlassian.net/browse/RUBY-259

- Adds form validation on temp_cards to not exceed 999
- Adds test coverage for details over the limit to prevent the form submit

### DIFF
--- a/app/views/waste_carriers_engine/cards_forms/new.html.erb
+++ b/app/views/waste_carriers_engine/cards_forms/new.html.erb
@@ -19,7 +19,12 @@
         <% end %>
 
         <%= f.label :temp_cards, t(".temp_cards_label"), class: "form-label" %>
-        <%= f.number_field :temp_cards, value: (@cards_form.temp_cards || 0), class: "form-control", min: 0, max: WasteCarriersEngine::CardsForm::MAX_TEMP_CARDS %>
+        <%= f.number_field :temp_cards,
+            value: (@cards_form.temp_cards || 0),
+            class: "form-control",
+            min: 0,
+            max: WasteCarriersEngine::CardsForm::MAX_TEMP_CARDS
+        %>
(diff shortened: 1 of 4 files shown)


app/views/waste_carriers_engine/cards_forms/new.html.erb
```diff
@@ -12,22 +12,24 @@
       <%= image_tag "example_card.jpg", alt: t(".card_alt_text") %>
     </div>
 
-    <% if @cards_form.errors[:temp_cards].any? %>
-    <div class="form-group form-group-error">
-    <% else %>
-    <div class="form-group">
-    <% end %>
+    <div class="form-group <%="form-group-error" if @cards_form.errors[:temp_cards].any? %>">
```
irisfaraway: I'm not opposed to the idea of this change as it does make the code more concise – however, as there are very similar lines across all our form views, I'd rather we tackled them all at once as a separate PR. What do you think?

cintamani: Sure why not :) The way I deal with this kind of refactoring is a "fix the file you are changing" approach. Mostly because finding the time for reviewing / refactoring them all at once is usually hard :D

irisfaraway: Totally get where you're coming from and we do that a lot as well (the Scout Rule!) I think this is one rare occasion where I disagree, just because I know that every form view has got at least one line that looks like this (and the same goes for waste exemptions). We're trying to focus on improving consistency throughout both apps at the moment as then we can do some much heavier refactoring down the line :)

cintamani: Makes sense :)

Issue #222, opened 2018-07-27
Rename repo to waste-carriers-engine


irisfaraway:
This matches the format of our other repositories (like flood-risk-engine) and would be clearer about what it's for.

PR #418, opened 2019-05-17
Update README after repo renaming

irisfaraway:
https://github.com/DEFRA/waste-carriers-engine/issues/222

We've finally renamed this repo from waste-carriers-renewals to waste-carriers-engine. This means the README needs to be updated to use the correct name.


irisfaraway:
Not sure why CodeClimate is hanging 🤔

PR #428, opened 2019-05-29
Tidy up views

cintamani:
The main goal is to improve the readability of our views

- Fix indentation across all views
- Refactor `if`s on the `error` form class addition to using ERB code inside the class value instead

Don't get scared by the length of this PR. The changes are 99% tabs and empty lines! :)


app/views/waste_carriers_engine/pdfs/certificate.html.erb
```diff
@@ -35,115 +35,113 @@
...
+                <tr>
+                  <th><%=t ".authority" %></th>
+                  <td></td>
+                </tr>
+              </thead>
+              <tbody>
+                <tr>
+                  <td class="table-column-text"><%=t ".name" %></td>
+                    <td>
+                      <%= wicked_pdf_image_tag("environment_agency_logo.png", {alt: "Environment Agency Logo", border: "0", height: "39", width: "129"}) %>
```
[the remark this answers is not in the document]
irisfaraway: No it's not

app/views/waste_carriers_engine/conviction_details_forms/new.html.erb
```diff
@@ -2,57 +2,54 @@
...
+        <%= f.submit t(".add_person_link"), class: "button-link" %>
+      </div>
+
+      <div class="form-group">
+        <%= f.submit t(".next_button"), class: "button" %>
+      </div>
+    <% end %>
+
+    <% if @conviction_details_form.number_of_existing_relevant_people > 0 %>
+      <div class="column-one-third">
```
irisfaraway: We need to close the div for `column-two-thirds` (starts on line 2) before we open this div. I think it is currently being closed on line 54 but we should move it up. I think we ran into similar issues on https://github.com/DEFRA/waste-exemptions-engine/pull/175

app/views/waste_carriers_engine/conviction_details_forms/new.html.erb
```diff
@@ -2,57 +2,54 @@
...
+        <h2 class="heading-small"><%= t(".list_of_people") %></h2>
+        <% @transient_registration.relevant_people.each do |person| %>
+          <ul class="key-person-list">
```
irisfaraway: Unrelated, but while we're in here... should we move `ul` outside this looping block? I think we did it for main people but not here!

app/views/waste_carriers_engine/conviction_details_forms/new.html.erb
cintamani: Oh! I just need to move that out! Good catch! :D

PR #458, opened 2019-08-20
Use shared business type validator

irisfaraway:
This PR removes our built-in business type validation in favour of the shared validator in defra-ruby-validators.


irisfaraway:
This breaks the test suite because we have separate logic for users who have said that they are overseas.

Planning to make some modifications to the gem to account for this.

irisfaraway:
https://github.com/DEFRA/defra-ruby-validators/pull/38 will sort this out (once it's approved, merged and released).

spec/support/shared_examples/validate_business_type.rb
```diff
@@ -2,33 +2,10 @@
...
-    context "when a business_type is not in the allowed list" do
-      before(:each) do
-        form.business_type = "foo"
-      end
-
-      it "is not valid" do
-        expect(form).to_not be_valid
-      end
-    end
+  it "validates the phone number using the BusinessTypeValidator class" do
```
Cruikshanks: Not quite sure its the phone number we are validating 😁 

```suggestion
  it "validates the business type using the BusinessTypeValidator class" do
```

irisfaraway: 🤦‍♀ Good spot!

PR #462, opened 2019-08-21
Use shared email validator

irisfaraway:
This PR removes our built-in email validation in favour of the shared validator in defra-ruby-validators.


irisfaraway:
CheckYourAnswersForm is starting to get pretty messy with all the I18n calls. Planning to clean this up in a separate PR.

PR #488, opened 2019-10-08
Forms refactoring - Remove extra reg_identifier param on submit

cintamani:

Remove extra reg_identifier parameter in the `submit` method. If necessary, it will be passed through the form parameters.


cintamani:
Codeclimate is a won't fix. Those will all be gone soon.

PR #490, opened 2019-10-08
Forms refactoring - Use strong parameters in Business type form

cintamani:

Use strong parameters on the business type form.
Fix temporary refactoring code which was not working when the method is defined in a child class.


app/controllers/waste_carriers_engine/forms_controller.rb
```diff
@@ -17,9 +17,9 @@ def create(form_class, form)
       return false unless set_up_form(form_class, form, params[form][:reg_identifier])
 
       # Submit the form by getting the instance variable we just set
-      if respond_to?(:transient_registration_attributes) # TODO: Temporary refactoring code.
+      begin # TODO: Temporary refactoring code.
```
cintamani:
```suggestion
      # TODO: Temporary refactoring code.
      begin
```

PR #493, opened 2019-10-09
Forms refactoring - Cards form

cintamani:

Use delegate and strong parameters in the Cards form. Update tests to match the fact that the params is casted to integer authomatically when passed to the model by Rails.


spec/requests/waste_carriers_engine/cards_forms_spec.rb
```diff
@@ -9,7 +9,7 @@ module WasteCarriersEngine
     include_examples "POST form",
                      "cards_form",
                      valid_params: { temp_cards: 2 },
-                     invalid_params: { temp_cards: "foo" },
+                     invalid_params: { temp_cards: 999999 },
```
cintamani:
```suggestion
                     invalid_params: { temp_cards: 999_999 },
```

PR #512, opened 2019-10-14
Forms Refactoring -Add ability to reference single document in collection

cintamani:

In Wex, using Active Record, we were able to simplify a lot our logic around addresses forms by redefining the association to the address as `has_one` associations.
Currently, there is no way to do the same with MongId 5.2, hence this adds some metaprogramming code that will allow a mode l to define a single object that is persisted inside a location and to override it.

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
+++ b/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
@@ -6,6 +6,7 @@ module WasteCarriersEngine
   module CanHaveRegistrationAttributes
     extend ActiveSupport::Concern
     include Mongoid::Document
+    include CanReferenceSingleDocumentInCollection
 
     # Rubocop sees a module as a block, and as such is not very forgiving in how
     # many lines it allows. In the case of this concern we have to list out all
@@ -18,7 +19,11 @@ module WasteCarriersEngine
       # the formatting we have in place.
       # rubocop:disable Metrics/LineLength
       embeds_many :addresses,               class_name: "WasteCarriersEngine::Address"
+      reference_one :contact_address, collection: :addresses, find_by: { address_type: "POSTAL" }
+      reference_one :registered_address, collection: :addresses, find_by: { address_type: "REGISTERED" }
+
       embeds_one :conviction_search_result, class_name: "WasteCarriersEngine::ConvictionSearchResult"
+
       embeds_many :conviction_sign_offs,    class_name: "WasteCarriersEngine::ConvictionSignOff"
       embeds_one :finance_details,          class_name: "WasteCarriersEngine::FinanceDetails", store_as: "financeDetails"
       embeds_many :key_people,              class_name: "WasteCarriersEngine::KeyPerson"
@@ -74,18 +79,6 @@ module WasteCarriersEngine
       field :title,                                            type: String
       field :total_fee,                                        type: String
 
-      def contact_address
-        return nil unless addresses.present?
-
-        addresses.where(address_type: "POSTAL").first
-      end
-
-      def registered_address
-        return nil unless addresses.present?
-
-        addresses.where(address_type: "REGISTERED").first
-      end
-
       def charity?
         business_type == "charity"
       end
@@ -102,14 +95,10 @@ module WasteCarriersEngine
       end
 
       def main_people
-        return [] unless key_people.present?
-
         key_people.where(person_type: "KEY")
       end
 
       def relevant_people
-        return [] unless key_people.present?
-
         key_people.where(person_type: "RELEVANT")
       end
 
--- a/app/models/concerns/waste_carriers_engine/can_reference_single_document_in_collection.rb
+++ b/app/models/concerns/waste_carriers_engine/can_reference_single_document_in_collection.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  module CanReferenceSingleDocumentInCollection
+    extend ActiveSupport::Concern
+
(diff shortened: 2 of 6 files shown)


app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -17,8 +17,12 @@ module CanHaveRegistrationAttributes
       # for comments in some places, and putting them on the line above breaks
       # the formatting we have in place.
       # rubocop:disable Metrics/LineLength
-      embeds_many :addresses,               class_name: "WasteCarriersEngine::Address"
+      embeds_many :addresses, class_name: "WasteCarriersEngine::Address"
+      reference_one :contact_address, collection: :addresses, find_by: { address_type: "POSTAL" }
+      reference_one :registered_address, collection: :addresses, find_by: { address_type: "REGISTERED" }
+
       embeds_one :conviction_search_result, class_name: "WasteCarriersEngine::ConvictionSearchResult"
+
```
cintamani:
```suggestion
```

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -109,14 +101,10 @@ def company_no_required?
       end
 
       def main_people
-        return [] unless key_people.present?
```
Cruikshanks: I need some context as to why this change has been made. Was it just housekeeping (deleting unnecessary code), or something related to the PR I don't understand???

I've spent some time invested in the Rails console both pre and post change trying to see for myself, and I can see it definitely makes a change to what gets returned when no `key_people` are present, but I don't think in any way that is meaningful to us.

So I guess I'd just like to understand better how it fits into the PR.

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -17,7 +18,10 @@ module CanHaveRegistrationAttributes
       # for comments in some places, and putting them on the line above breaks
       # the formatting we have in place.
       # rubocop:disable Metrics/LineLength
-      embeds_many :addresses,               class_name: "WasteCarriersEngine::Address"
+      embeds_many :addresses, class_name: "WasteCarriersEngine::Address"
+      reference_one :contact_address, collection: :addresses, find_by: { address_type: "POSTAL" }
```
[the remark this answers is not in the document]
cintamani:
```suggestion
      # This is our own custom association. See CanReferenceSingleDocumentInCollection for details
      reference_one :contact_address, collection: :addresses, find_by: { address_type: "POSTAL" }
```

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
cintamani: Just me opening the class and thinking we don't need that as `has_many` association should return an empty array even after a where is applied even if empty. Is this not the case for MongoId?

Cruikshanks: Sort of. With that line removed it always returns a [Mongoid::Criteria](https://www.rubydoc.info/github/mongoid/mongoid/Mongoid/Criteria) which you then can interact with like an array. So for example if `key_people` does not exist `my_reg.main_people.length` returns 0 whether the line is there or not.

Which is why I said it's probably fine and not going to break anything, but perhaps in future use commits to separate out what is the feature, and what is other bits like housekeeping, fixes, code formatting changes etc. I just wasn't sure if the single document stuff relied on this change because of something I wasn't seeing.

app/models/concerns/waste_carriers_engine/can_reference_single_document_in_collection.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  module CanReferenceSingleDocumentInCollection
```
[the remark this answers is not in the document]
Cruikshanks: Just that additional documentation/comment here @cintamani and then happy to approve!

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
cintamani: I understand! Let me scrap this change out of this PR :)

app/models/concerns/waste_carriers_engine/can_reference_single_document_in_collection.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+# This module's aim is to implement a way to reference documetns in collection but singularly
+# so that they can be treated as `has_one` associations. Currently on the WEX project we achieve the same
+# functionality thanks to ActiveRecord Relation ability of specify custom associations using default scopes.
+# But with the version of MongoId we use there is no ability of doing so.
```
Cruikshanks:
```suggestion
# This module's aim is to implement a way to reference a single document in a
# collection so that they can then be treated as `has_one` associations.
# In projects using ActiveRecord like WEX we achieve the same functionality
# thanks to ActiveRecord's Relation ability to specify custom associations using
# default scopes. But because here we are using an old version of MongoDB, we
# are also stuck with a version of Mongoid which does not have this ability.
```

Cruikshanks: Turns out multiline comments don't apply to multiline suggestions ☹️

cintamani: yep it does not! Shame! :D

PR #513, opened 2019-10-15
Use Fetch instead of Require when fetching parameters

cintamani:
Given that: https://github.com/rails/strong_parameters/issues/162 is an issue for us, as we want to allow people to submit empty forms, this changes the strategy of fetching parameters from `require` to `fetch` as suggested by the official Rails documentation: https://guides.rubyonrails.org/action_controller_overview.html#more-examples

### DIFF
--- a/spec/support/shared_examples/post_form.rb
+++ b/spec/support/shared_examples/post_form.rb
@@ -96,9 +96,11 @@ RSpec.shared_examples "POST form" do |form, options|
 
         context "when the params are empty" do
           it "does not throw an error" do
+            # rubocop:disable Style/BlockDelimiters
             expect {
               post_with_params(form, { reg_identifier: transient_registration.reg_identifier })
             }.not_to raise_error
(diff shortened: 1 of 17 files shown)


spec/support/shared_examples/post_form.rb
```diff
@@ -94,6 +94,14 @@
           end
         end
 
+        context "when the params are empty" do
+          it "does not throw an error" do
+            expect {
```
cintamani:
```suggestion
            # rubocop:disable Style/BlockDelimiters
            expect {
```

spec/support/shared_examples/post_form.rb
```diff
@@ -94,6 +94,14 @@
...
+              post_with_params(form, { reg_identifier: transient_registration.reg_identifier })
+            }.not_to raise_error
```
cintamani:
```suggestion
            }.not_to raise_error
            # rubocop:enable Style/BlockDelimiters
```

PR #514, opened 2019-10-17
Forms Refactoring - Address lookup forms

cintamani:

Apply refactoring to Lookup forms.
The forms are now sending a `uprn` through the select field, which is then used to match the correct address with the list of temp_addresses from OS places. This ensure we can use standard Rails strategy of Strong Parameters and mass assign.
Since the assignment of contact and company addresses is done through a non-standard rails association, what is different from WEX is that in this case we will pass to the `transient_registration` an instance of `Address` rather than `address_attributes`. This is due to the fact that we cannot use `accept_nested_params` for a fake association.


app/forms/waste_carriers_engine/company_address_form.rb
```diff
@@ -1,32 +1,20 @@
...
-    def temp_postcode
-      temp_company_postcode
-    end
-
-    def saved_address
-      transient_registration.registered_address
-    end
+    def submit(params)
+      company_address_params = params.fetch(:company_address, {})
+      # binding.pry
```
cintamani:
```suggestion
```

PR #519, opened 2019-10-18
Forms refactoring: Address Manual forms

cintamani:

Apply strong parameters strategy to address manual forms.


cintamani:
Codeclimate is a won't fix

PR #521, opened 2019-10-20
Add registration.can_start_renewal?

irisfaraway:
This was suggested in reviews of https://github.com/DEFRA/waste-carriers-back-office/pull/385

I opted for `can_start_renewal?` instead of `renewable?` as the conditions for starting and completing a renewal are different, and this seemed like a clearer name.


irisfaraway:
Could be tweaked if https://github.com/DEFRA/waste-carriers-engine/pull/520 is merged first.

PR #526, opened 2019-10-21
Forms refactoring - Create concern for Cannot submit

cintamani:

Create a concern to deal with unsubmittable forms


app/forms/concerns/waste_carriers_engine/cannot_submit.rb
```diff
@@ -0,0 +1,10 @@
+module WasteCarriersEngine
```
cintamani:
```suggestion
# frozen_string_literal: true.

module WasteCarriersEngine
```

app/forms/concerns/waste_carriers_engine/cannot_submit.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true.
```
cintamani:
```suggestion
# frozen_string_literal: true
```

PR #527, opened 2019-10-22
Forms refactoring - People forms controllers

cintamani:

Use strong parameters in people forms


cintamani:
Codeclimate is a won't fix

PR #530, opened 2019-10-22
Forms refactoring - Remove temp code in BaseForm

cintamani:
From https://eaflood.atlassian.net/browse/RUBY-693

Remove temporary code hack since now all forms implement strong parameters


app/controllers/waste_carriers_engine/forms_controller.rb
```diff
@@ -34,6 +28,12 @@ def go_back
 
     private
 
+    def transient_registration_attributes
+      # Default behavuour - permit no params
+      # Override in subclasses if needed
```
cintamani:
```suggestion
      # Override in subclasses when needed
```

PR #533, opened 2019-10-23
Move setting up of ASSISTED_DIGITAL flag in AASM machine

cintamani:
Solve: https://eaflood.atlassian.net/browse/RUBY-718

We used the `RenewalCompletionService` to setup the metaData route of the transient_registration to the correct application-level configuration of `ASSISTED_DIGITAL` vs `DIGITAL`.
This was causing some registration to be incorrectly flagged as `ASSISTED_DIGITAL`, because the same service is run in the back-office application when a conviction check or payment check have been completed.
In order to make sure that we only flag the registration at the end of the forms workflow, we have moved the setting from the Service to the workflow AASM state machine.

### DIFF
--- a/app/services/waste_carriers_engine/renewal_completion_service.rb
+++ b/app/services/waste_carriers_engine/renewal_completion_service.rb
@@ -54,7 +54,7 @@ module WasteCarriersEngine
     end
 
     def update_meta_data
-      @registration.metaData.route = Rails.configuration.metadata_route
+      @registration.metaData.route = @transient_registration.metaData.route
       @registration.metaData.renew
       @registration.metaData.date_registered = Time.now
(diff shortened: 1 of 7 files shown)


app/services/waste_carriers_engine/renewal_completion_service.rb
```diff
@@ -54,7 +54,7 @@ def update_registration
     end
 
     def update_meta_data
-      @registration.metaData.route = Rails.configuration.metadata_route
+      @registration.metaData.route = @transient_registration.metaData&.route || Rails.configuration.metadata_route
```
cintamani:
```suggestion
      @registration.metaData.route = @transient_registration.metaData.route
```

PR #534, opened 2019-10-23
Add payment check methods to registration

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-704

This will let us check for unpaid balances on registrations as well as renewals.

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
+++ b/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
@@ -157,6 +157,10 @@ module WasteCarriersEngine
       def declaration_confirmed?
         declaration == 1
       end
+
+      def unpaid_balance?
+        finance_details.present? && finance_details.balance.positive?
+      end
     end
     # rubocop:enable Metrics/BlockLength
   end
--- a/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
+++ b/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
@@ -159,7 +159,9 @@ module WasteCarriersEngine
       end
 
       def unpaid_balance?
-        finance_details.present? && finance_details.balance.positive?
+        return true if finance_details.presence&.balance&.positive?
+
+        false
       end
     end
     # rubocop:enable Metrics/BlockLength
--- a/app/models/waste_carriers_engine/registration.rb
+++ b/app/models/waste_carriers_engine/registration.rb
@@ -26,12 +26,10 @@ module WasteCarriersEngine
     validates :tier,
               inclusion: { in: %w[UPPER LOWER] }
 
+    alias pending_payment? unpaid_balance?
+
     def can_start_renewal?
       renewable_tier? && renewable_status? && renewable_date?
-    end
-
-    def pending_payment?
-      unpaid_balance?
     end
 
     private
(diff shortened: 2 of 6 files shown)


app/models/waste_carriers_engine/registration.rb
```diff
@@ -30,6 +30,10 @@ def can_start_renewal?
       renewable_tier? && renewable_status? && renewable_date?
     end
 
+    def pending_payment?
+      unpaid_balance?
```
cintamani: can we use `alias` here?

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -157,6 +157,10 @@ def update_last_modified
       def declaration_confirmed?
         declaration == 1
       end
+
+      def unpaid_balance?
+        finance_details.present? && finance_details.balance.positive?
```
cintamani: [NIT]
```suggestion
        finance_details.presence&.balance.positive?
```

spec/models/waste_carriers_engine/registration_spec.rb
```diff
@@ -566,6 +566,40 @@ module WasteCarriersEngine
       end
     end
 
+    describe "#pending_payment?" do
```
cintamani: if we `alias`, we won't need to test this, instead we should test `unpaid_balance?`. Also, this test is not a pure unit, shall we still go for it?

irisfaraway:
Updated after review comments.

I realise that the "Can have registration attributes" shared examples are currently only testing transient_registrations 🤦‍♀ I'm fixing this in another branch at the moment though so hopefully it won't be a blocker to merging this.

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -157,6 +157,12 @@ def update_last_modified
       def declaration_confirmed?
         declaration == 1
       end
+
+      def unpaid_balance?
+        return true if finance_details.presence&.balance&.positive?
```
cintamani: Just a question: the previous version of the code always assumed a finance_details, if present, should always have a `balance` present that has a numeric value/object. Is this not the case any more?

app/models/waste_carriers_engine/transient_registration.rb
```diff
@@ -102,7 +102,7 @@ def renewal_application_submitted?
     end
 
     def pending_payment?
```
cintamani: I was expecting to see the spec for this method to change as well with an `expect(subject).to receive(:unpaid_balance?) ....`, shall we update it?

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
irisfaraway: It should have a `balance` (although MongoDB is less strict about this sort of thing). But Rubocop kicked off when I had `finance_details.presence&.balance.positive?` – I think it doesn't like to have `.` calls chained after `&.` calls.

cintamani: Thanks for the explanation! :)

app/models/waste_carriers_engine/transient_registration.rb
irisfaraway: Fixed: https://github.com/DEFRA/waste-carriers-engine/pull/534/commits/b4d267b38f72e90177cfd7b70fb39df567b932d7

PR #535, opened 2019-10-23
Add pending_manual_conviction_check? method to registration

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-704

This may seem redundant, but it means that registrations and transient_registrations have the same interface, which will simplify the checks for the back office dashboard.

### DIFF
--- a/app/models/waste_carriers_engine/registration.rb
+++ b/app/models/waste_carriers_engine/registration.rb
@@ -26,12 +26,10 @@ module WasteCarriersEngine
     validates :tier,
               inclusion: { in: %w[UPPER LOWER] }
 
+    alias pending_manual_conviction_check? conviction_check_required?
+
     def can_start_renewal?
       renewable_tier? && renewable_status? && renewable_date?
-    end
-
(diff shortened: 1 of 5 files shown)


app/models/waste_carriers_engine/registration.rb
```diff
@@ -30,6 +30,10 @@ def can_start_renewal?
       renewable_tier? && renewable_status? && renewable_date?
     end
 
+    def pending_manual_conviction_check?
+      conviction_check_required?
```
cintamani: If we `alias` we don't even need to unit test this :D

spec/models/waste_carriers_engine/registration_spec.rb
```diff
@@ -566,26 +566,6 @@ module WasteCarriersEngine
       end
     end
 
-    describe "#pending_manual_conviction_check?" do
```
cintamani: I think we should add test coverage for the `conviction_check_required?` implementation in `CanHaveRegistrationAttribute` then :D

PR #540, opened 2019-10-27
[FOR DISCUSSION] Specs refactoring - Unit tests on services

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-692

This is a very simple example of using Double rather than real objects. Not much changes from the real thing really. I have started other forms refactoring of specs too but I gave up as most of them are too tedious and time consuming due to the number of operations they perform. And since this is not any urgent work, I decided to not go further.


cintamani:
Travis has some issues with a test not related to this

PR #548, opened 2019-11-07
Create model for conviction entities

irisfaraway:

This is the first step of implementing conviction checking entirely in the engine instead of making calls to waste-carriers-service.

This PR sets up an Entity model to match the data we store in the database.

### DIFF
--- a/app/models/waste_carriers_engine/convictions_check/entity.rb
+++ b/app/models/waste_carriers_engine/convictions_check/entity.rb
@@ -19,9 +19,11 @@ module WasteCarriersEngine
       field :incidentNumber, as: :incident_number, type: String
 
       scope :matching_organisation_name, lambda { |term|
-        term = ::Regexp.escape(term) if term.present?
+        escaped_term = ::Regexp.escape(term) if term.present?
+        # If the name ends with a full stop, treat that as optional in the regex
+        term_with_optional_trailing_full_stop = escaped_term.gsub(/\.$/, ".?") if escaped_term.present?
 
-        where(name: /#{term}/i)
+        where(name: /#{term_with_optional_trailing_full_stop}/i)
       }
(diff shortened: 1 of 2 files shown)


irisfaraway:
Aside from the question of whether or not to exclude certain terms (see https://github.com/DEFRA/waste-carriers-service/blob/398abf7b35d953a3878d6490efb80f6b68b1cdb9/src/main/java/uk/gov/ea/wastecarrier/services/match/CompanyMatch.java#L24-L33) – this is ready for review.

irisfaraway:
Resolved the question about the common terms on Slack, so bumping this back out of review to sort it.

app/models/waste_carriers_engine/convictions_check/entity.rb
```diff
@@ -0,0 +1,83 @@
...
+
+        where(name: /#{escaped_first_name}/i).and(name: /#{escaped_last_name}/i)
+      }
+
+      scope :matching_date_of_birth, lambda { |term|
+        where(date_of_birth: term)
+      }
+
+      scope :matching_company_number, lambda { |term|
+        escaped_term = ::Regexp.escape(term) if term.present?
```
cintamani: shall we return if there is no term present? Line 41 will fail if the term is not present and we are missing test coverage for when a term is empty

irisfaraway: Good shout – I've added test coverage and set the scopes to raise an error if the terms aren't present!

irisfaraway:
Don't know why CodeClimate is stuck but this is ready for review again 🤷‍♀

irisfaraway:
@cintamani Cool. I can crack on with other stuff in the meantime so let's sort it out and merge then :)

PR #550, opened 2019-11-08
Add RenewingRegistration model

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-756

In order to align the codebase with WEX we want to make TransientRegistration a STI model.
This PR includes the new model that will split renewing functionality out of TransientRegistration into RenewingRegistration model.

The class is a copy-paste of most the code in `TransientRegistration`


cintamani:
Codeclimate is a won't fix

PR #554, opened 2019-11-08
Set up PersonMatchService for individual conviction checks

irisfaraway:

This service will be used to check for matches on individual people.

---

Depends on https://github.com/DEFRA/waste-carriers-engine/pull/548


irisfaraway:
Basically ready for review but will need re-approval after #548 is merged.

spec/services/waste_carriers_engine/convictions_check/person_match_service_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+                 name: "baz")
+        end
+        let(:entity_b) do
+          double(:entity,
+                 system_flag: "qux",
+                 incident_number: "quux",
+                 name: "quuz")
+        end
+
+        let(:subject) do
```
cintamani: Why `let`?
```suggestion
        subject do
```
Or, since the `subject` is actually the `result` of the `run` action, can we call this `result`? It will avoid confusion with the fact that `subject` is expected to contain an instance of the testing class by convention ::)

PR #555, opened 2019-11-08
Set up OrganisationMatchService for org-level conviction checks

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-741

This service will be used to check for matches on the registering business or organisation.

---

Depends on https://github.com/DEFRA/waste-carriers-engine/pull/554

PR #556, opened 2019-11-08
Update EntityMatchingService to call new convictions check

irisfaraway:

Java be gone! This PR swaps out our old calls to the Java service in favour of calling our new conviction match services instead.

Because the data we are returning is slightly nicer (no weird Unix time in seconds), there may be some knock-on effects in other classes too.

---

Depends on https://github.com/DEFRA/waste-carriers-engine/pull/555


irisfaraway:
Basically ready for review but will need re-approval after #555  is merged.

app/models/waste_carriers_engine/conviction_search_result.rb
```diff
@@ -20,13 +20,12 @@ class ConvictionSearchResult
...
-      result.matched_name = data["matched_name"]
-      # The service provides searched_at as milliseconds from the Unix epoch, so conversion is required
-      result.searched_at = Time.at(data["searched_at"] / 1_000).to_datetime
-      result.confirmed = data["confirmed"]
+      result.match_result = data[:match_result]
+      result.matching_system = data[:matching_system]
+      result.reference = data[:reference]
+      result.matched_name = data[:matched_name]
+      result.searched_at = data[:searched_at]
+      result.confirmed = data[:confirmed]
```
cintamani: NIT: https://apidock.com/rails/v4.2.1/Hash/slice seems to be useful here :)

PR #558, opened 2019-11-12
Add rejected method to transient registration

cintamani:
This adds a method on the transient registration to check for a rejected status from the Metadata


cintamani:
Oh! Nothing! I didn't spot the method there 🤦‍♂ will just add an `alias`!

PR #559, opened 2019-11-12
Add rejected method to registrations

cintamani:
When a transient registration is in a `Revoked` status, due to conviction checks not passing, we consider it at all effects a `Rejected` registration. In order to improve readibility of the code there, I want to have a method that is called `rejected` rather than `revoked`.


Cruikshanks:
@irisfaraway is right. We can 'reject' a transient registration because it's not yet registered.

However, once it becomes a registration, it is active and live. So all we can do at this point is 'revoke' it.

Agree 'revoke' on an in-progress registration doesn't make much sense, but 'reject' on registration is an incorrect term so we couldn't use it there.

Does this feedback help @cintamani ?

cintamani:
I get you :) Will update

PR #563, opened 2019-11-18
Add lower and upper tier check methods to Registrations

cintamani:
I am not aware of other code implementing those, and currently I have the need to distinguish between a `Lower` and `Upper` tier registration, hence I am adding this in.

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
+++ b/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
@@ -103,6 +103,14 @@ module WasteCarriersEngine
         business_type == "overseas"
       end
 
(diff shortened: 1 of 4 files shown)


cintamani:
@Cruikshanks I have thought that too. But I have decided that this is a question to answer later when we work on the front-end workflow. I was thinking that since we don't set the value until later in the journey, we should think it better when we put it there. 
However, since you are pointing at it, and the implementation will not create any error if the value is `nil`, I'll move it into the shared concern :)

irisfaraway:
I'd agree with moving them to the shared concern as well.

cintamani:
I was so sure I did push the change yesterday! Instead, I didn't! Sorry for the spam!

PR #565, opened 2019-11-19
Treat punctuation as optional for org name conviction checks

irisfaraway:

This fixes an issue spotted in QA. Punctuation should be treated as optional when matching organisation names - for example, if a user enters "Waste Not (Want Not) UK" that should still match a record called "Waste Not Want Not UK".

### DIFF
--- a/app/models/waste_carriers_engine/convictions_check/entity.rb
+++ b/app/models/waste_carriers_engine/convictions_check/entity.rb
@@ -81,15 +81,34 @@ module WasteCarriersEngine
       private_class_method def self.org_name_search_term(term)
         return if term.blank?
 
-        # Trim trailing full stops
-        term_without_trailing_full_stops = term.gsub(/\.$/, "")
+        # The steps for processing the org name must be done in this order:
 
-        # Remove the words we want to ignore
-        word_array = term_without_trailing_full_stops.downcase.split(" ")
+        # 1. Remove trailing full stops
+        without_full_stops = term.gsub(/\.$/, "")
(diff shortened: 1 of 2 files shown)


app/models/waste_carriers_engine/convictions_check/entity.rb
```diff
@@ -88,15 +88,34 @@ def self.matching_organisations(name:, company_no: nil)
...
-        word_array = term_without_trailing_full_stops.downcase.split(" ")
+      private_class_method def self.term_without_ignorable_org_words(term)
+        word_array = term.downcase.split(" ")
         word_array.reject! { |word| IGNORABLE_ORG_NAME_WORDS.include?(word) }
-        term_without_ignorable_words = word_array.join(" ")
+        word_array.join(" ")
+      end
+
+      private_class_method def self.term_with_optional_punctuation(term)
+        # These are characters we want to treat as optional
```
cintamani: Shall we reflect this in the variable's name? Like: `optional_chars` instead of `punctuation_chars` :)

Cruikshanks: That probably makes sense. I got what was going on without issue so didn't spot this, but now that you highlight it, I see why.

irisfaraway: @cintamani Good shout! Updated: https://github.com/DEFRA/waste-carriers-engine/pull/565/commits/4c1852a5a725b3c324beae42ca40f57f5970e4bf

PR #567, opened 2019-11-20
Add renewable? method to all registrations

cintamani:
This adds the `renewable?` method to all types of registrations, as per WEX.

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
+++ b/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
@@ -124,6 +124,10 @@ module WasteCarriersEngine
         conviction_sign_offs.last.rejected?
       end
 
+      def renewable?
+        false
+      end
+
       def main_people
         return [] unless key_people.present?
 
--- a/app/models/waste_carriers_engine/registration.rb
+++ b/app/models/waste_carriers_engine/registration.rb
@@ -33,6 +33,8 @@ module WasteCarriersEngine
       renewable_tier? && renewable_status? && renewable_date?
     end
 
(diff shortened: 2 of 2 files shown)


app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -124,6 +124,10 @@ def rejected_conviction_checks?
         conviction_sign_offs.last.rejected?
       end
 
+      def renewable?
```
Cruikshanks: Why are we adding this here? Only 'registrations' can be renewed, so I'm not sure why it would make sense to also apply this to a renewing registration, and in future an edit registration.

app/models/waste_carriers_engine/registration.rb
```diff
@@ -33,6 +33,8 @@ def can_start_renewal?
       renewable_tier? && renewable_status? && renewable_date?
     end
 
+    alias renewable? can_start_renewal?
```
Cruikshanks: I appreciate having the alias will save some work now. But if `renewable?` is the name we should be using (because its clearer and consistent with WEX) wouldn't it be better to bite the bullet and rename `can_start_renewal?` instead?

N.B. Where in WEX is [renewable](https://github.com/DEFRA/waste-exemptions-engine/search?q=renewable&unscoped_q=renewable). I can't find it?

cintamani: I got confused with the transient registration `renewal?` method 🤦‍♂ 
I have a lot of places right now in the back-office where I have a registration (which can be transit or not) and I need to check if is renewable. I will actually do that in the presenter for now, as looking back here is not such a good idea

PR #570, opened 2019-11-29
Add service to order copy cards

cintamani:

Add a reusable service that given a registration, a user, a number of copy cards and a user, generates a copy cards only order

### DIFF
--- a/app/models/waste_carriers_engine/order.rb
+++ b/app/models/waste_carriers_engine/order.rb
@@ -25,18 +25,11 @@ module WasteCarriersEngine
     field :manualOrder, as: :manual_order,           type: String
     field :order_item_reference,                     type: String
 
+    # TODO: Move to a service
     def self.new_order(transient_registration, method, current_user)
-      order = Order.new
+      order = new_order_for(current_user)
 
       card_count = transient_registration.temp_cards
-
-      order[:order_id] = order.generate_id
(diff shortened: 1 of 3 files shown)


app/models/waste_carriers_engine/order.rb
```diff
@@ -25,19 +25,12 @@ class Order
     field :manualOrder, as: :manual_order,           type: String
     field :order_item_reference,                     type: String
 
+    # TODO: Move to a service
```
Cruikshanks: If the intention is to move this to an `OrderRenewalService` (or something like that) would it be better to do that refactoring first and then add your new service?

As is I'm just not sure I like the idea that if we never get round to this `TODO` we'll have `Order` create the order with `new_order`, set the order items as if its a renewal, but then `OrderAdditionalCardsService` take that order and simply overwrites the array. Plus we're creating orders in 2 different ways.

cintamani: This service do not use the code in `new_order` so I have left the TODO for the next person touching it. We should get there naturally when we work out the front end flow.

Cruikshanks: 🤦‍♂ Yep, was pointed out to me I was misreading the changes!

cintamani:
I am pretty sure I actually need this as it is now, so I am merging this in

PR #573, opened 2019-12-02
Add OrderCopyCardsRegistration model

cintamani:

Add an OrderCopyCardsRegistration model. This is a new type of TransientRegistration which will be used for the ordering of extra copy cards flow on active registrations.


spec/factories/order_copy_cards_registration.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
```
irisfaraway: Do we need all this stuff in the factory - overseas address options, etc?

spec/models/waste_carriers_engine/order_copy_cards_registration_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    subject(:order_copy_cards_registration) { build(:order_copy_cards_registration, :has_required_data) }
+
+    describe "workflow_state" do
+      context "when a OrderCopyCardsRegistration is created" do
+        it "has the state :copy_cards_form" do
+          expect(order_copy_cards_registration).to have_state(:copy_cards_form)
+        end
+      end
+
+      context "transitions" do
```
irisfaraway: For WCR renewing regs and for all the WEX journeys, we kept the transition specs in separate files (see https://github.com/DEFRA/waste-exemptions-engine/tree/master/spec/models/waste_exemptions_engine) - should we do that here as well for consistency? I appreciate that the journey here is much simpler, but I think it would be helpful so that we always know where to look for workflow specs.

irisfaraway: WEX had a lot of shared examples for testing transitions as well - may be worth looking if those would be useful here too.

spec/models/waste_carriers_engine/order_copy_cards_registration_spec.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module WasteCarriersEngine
+  RSpec.describe OrderCopyCardsRegistration, type: :model do
+    subject(:order_copy_cards_registration) { build(:order_copy_cards_registration, :has_required_data) }
```
irisfaraway: I thought we weren't using factories in our unit tests?

spec/factories/order_copy_cards_registration.rb
cintamani: oh shut! No we don't! I forgot to delete XD

spec/models/waste_carriers_engine/order_copy_cards_registration_spec.rb
cintamani: will do :)

spec/models/waste_carriers_engine/order_copy_cards_registration_spec.rb
cintamani: I don't think we can apply it to models stuff like validations and the workflow can't really be tested with doubles

cintamani: plus this is a build so litterally is the same as `described_class.new` in this case

cintamani:
Yes we can't reuse the payment form, I am aware of that. But my plan is to rename it when I get to it (=build the form)

cintamani:
@irisfaraway this should be good now :)

PR #576, opened 2019-12-04
Add copy cards main forms

cintamani:

This add the CopyCardForm and the CopyCardsPaymentForm.
Both Bank transfer and Worldpay forms needs mor thinking, so in order to facilitate review this only contains the other forms.
View implementation is a TODO.

### DIFF
--- a/app/forms/waste_carriers_engine/copy_cards_payment_form.rb
+++ b/app/forms/waste_carriers_engine/copy_cards_payment_form.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class CopyCardsPaymentForm < CardsForm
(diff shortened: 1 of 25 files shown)


app/forms/waste_carriers_engine/copy_cards_payment_form.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class CopyCardsPaymentForm < CardsForm
```
cintamani:
```suggestion
  class CopyCardsPaymentForm < BaseForm
```

cintamani:
Sorry for the spam! I thought something was missing but I was misled by my local git 🤦‍♀

PR #580, opened 2019-12-05
Add service that generates Base(58) secure tokens

Cruikshanks:
In WEX and FRAE we have relied on using [HasSecureToken](https://github.com/robertomiranda/has_secure_token) to generate unique random tokens of a fixed length in the objects we create.

We have a need to unqiuely identify a record, but cannot use its ID because it also has to be something 'unguessable'.

HasSecureToken works for WEX and FRAE because it is designed to work with ActiveRecord models. WCR on the overhand uses MongoDb, which means we can't use it.

So instead, we've copied it wholesale into our project as a service we can call when we need to generate a token. Thanks @robertomiranda 😁!

### DIFF
--- a/app/services/waste_carriers_engine/secure_token_service.rb
+++ b/app/services/waste_carriers_engine/secure_token_service.rb
@@ -3,7 +3,6 @@
 require "securerandom"
 
 module WasteCarriersEngine
-
   # SecureTokenService generates a random base58 string of length 24.
   #
   # SecureRandom::base58 is used to generate the 24-character unique tokens, so
--- a/app/services/waste_carriers_engine/secure_token_service.rb
+++ b/app/services/waste_carriers_engine/secure_token_service.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "securerandom"
+
+module WasteCarriersEngine
+
+  # SecureTokenService generates a random base58 string of length 24.
+  #
+  # SecureRandom::base58 is used to generate the 24-character unique tokens, so
+  # collisions are highly unlikely.
+  #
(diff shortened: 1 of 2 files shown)


spec/services/waste_carriers_engine/secure_token_service_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        10.times do
+          latest = described_class.run
+
+          expect(results).not_to include(latest)
+
+          results.push(latest)
+        end
+      end
+    end
+
```
cintamani: Super-NIT:
```suggestion
```

app/services/waste_carriers_engine/secure_token_service.rb
```diff
@@ -0,0 +1,31 @@
...
+    BASE58_ALPHABET = ("0".."9").to_a + ("A".."Z").to_a + ("a".."z").to_a - %w[0 O I l]
+
+    def run
+      SecureRandom.random_bytes(24).unpack("C*").map do |byte|
+        idx = byte % 64
+        idx = SecureRandom.random_number(58) if idx >= 58
+        BASE58_ALPHABET[idx]
+      end.join
+    end
+
```
cintamani: Super-NIT:
```suggestion
```

app/services/waste_carriers_engine/secure_token_service.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require "securerandom"
+
+module WasteCarriersEngine
+
```
cintamani: Super-NIT:
```suggestion
```

PR #583, opened 2019-12-06
Set up RegistrationCompletionService

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-793

When conviction checks for new registrations are moved into the back office, this is the first time we'll need to 'complete' a new registration in the new codebase. So this PR creates a RegistrationCompletionService to handle that.

When the new registration journey starts to use transient_registrations, this will probably need some modifying to deal with those instead of the registration model.

### DIFF
--- a/app/services/waste_carriers_engine/registration_completion_service.rb
+++ b/app/services/waste_carriers_engine/registration_completion_service.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class RegistrationCompletionService < BaseService
+    def run(registration:)
+      registration.metaData.date_activated = Time.current
+      registration.metaData.activate!
+    end
+  end
+end
--- a/app/services/waste_carriers_engine/registration_completion_service.rb
+++ b/app/services/waste_carriers_engine/registration_completion_service.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 module WasteCarriersEngine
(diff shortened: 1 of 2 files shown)


app/services/waste_carriers_engine/registration_completion_service.rb
```diff
@@ -0,0 +1,46 @@
...
+    def no_pending_conviction_check?
+      if @registration.pending_manual_conviction_check?
+        raise_convictions_check_error
+      else
+        true
+      end
+    end
+
+    def raise_unpaid_balance_error
+      raise "Registration #{@registration.reg_identifier} cannot be activated due to unpaid balance"
```
cintamani: Can we create actual error classes to raise here?

app/services/waste_carriers_engine/registration_completion_service.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def raise_convictions_check_error
+      raise "Registration #{@registration.reg_identifier} cannot be activated due to pending convictions check"
```
cintamani: Ditto

app/services/waste_carriers_engine/registration_completion_service.rb
irisfaraway: Done: https://github.com/DEFRA/waste-carriers-engine/pull/583/commits/fb9bd9830e4684f58816367d5b0fa03c83a31f22

PR #584, opened 2019-12-06
Persist setting of expiries_on when activating

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-793

Spotted that while the `activate` event sets the `expires_on` in memory, it doesn't persist it. This means that if the registration was assigned an `expires_on` when it was first submitted through the frontend, that value sticks, although the behaviour we want is for it to expire 3 years after approval.

So this fixes that.

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
+++ b/app/models/concerns/waste_carriers_engine/can_change_registration_status.rb
@@ -57,7 +57,7 @@ module WasteCarriersEngine
 
       # Transition effects
       def set_expiry_date
-        registration.expires_on = Rails.configuration.expires_after.years.from_now
+        registration.update_attributes(expires_on: Rails.configuration.expires_after.years.from_now)
       end
 
       def log_status_change
--- a/spec/models/waste_carriers_engine/registration_spec.rb
+++ b/spec/models/waste_carriers_engine/registration_spec.rb
@@ -393,13 +393,17 @@ module WasteCarriersEngine
         end
 
         context "when a registration is activated" do
-          let(:registration) { build(:registration, :is_pending) }
+          let(:registration) { create(:registration, :has_required_data, :is_pending) }
+
+          before { allow(Rails.configuration).to receive(:expires_after).and_return(3) }
 
           it "sets expires_on 3 years in the future" do
             expect(registration.expires_on).to be_nil
-            registration.metaData.activate
+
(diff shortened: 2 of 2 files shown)


Cruikshanks:
Sorry, `request changes` was the best way I could think to flag this might need further checking.

I'm looking here at what the old code does (I might have the wrong place) and I can't see where it updates the `expires_on` value https://github.com/DEFRA/waste-carriers-frontend/blob/master/app/controllers/registrations_controller.rb#L873

From what I can see digging through the old code, let's assume the following

- start a new registration and `expires_on` is immediately set
- registration gets flagged and you choose to pay by BACS

Based on this I believe

- if payment is entered first the `expires_on` does not change because the registration still needs approval. When approved `expires_on` remains as when first set
- if approval is done first, `expires_on` is not changed. But when the payment is then entered [waste-carriers-service](https://github.com/DEFRA/waste-carriers-service) sees the registration can be activated (and does so) including setting the `expires_on` to the current date

https://github.com/DEFRA/waste-carriers-service/blob/master/src/main/java/uk/gov/ea/wastecarrier/services/resources/PaymentResource.java#L70
https://github.com/DEFRA/waste-carriers-service/blob/master/src/main/java/uk/gov/ea/wastecarrier/services/helper/PaymentHelper.java#L118

In a weird way, I think this tallies with what NCCC have stated previously (if a bit broken). We need to check the following with NCCC but my understanding is

- a registration starts from the date it is paid for, irrespective of there being outstanding checks. So the `expires_on` of a PENDING registration should be updated to match the date the balance is updated to 0
- approving conviction checks should not change when the registration starts (and therefore expires)
- NCCC only carry out conviction checks on paid registrations, so our second scenario above would never happen, which means all current registrations will have the `expires_on` they were given when initialised

If my assumptions are correct, the bit that's currently broken is waste-carriers-service should be updating the `expires_on` when payment is received even if the registration is not ready to be activated

So TLDR, I think we just need to have a quick call with NCCC to confirm some of this stuff.

I now await the 'actually line # 234 does it, which is why I made this PR. But thanks for the waffle anyway'. 😁

irisfaraway:
@Cruikshanks I'm not sure where in the code it is, or what the intended behaviour is – but the date being updated is what seemed to happen when I tested the behaviour of the old app! I was also confused but figured I should replicate what we already had.

A check-in with NCCC sounds like it would be useful though.

PR #586, opened 2019-12-09
[WIP - Splitting ]Add CopyCardsBankTransferForm to the order copy cards flow

cintamani:

This adds the copy cards bank transfer form functionality.
Part of this work includes the creation of a completion service for the flow. To achieve this, common code from the Renewal completion service have been externalised to a concern.
This also lead to the rename of the completion state in the worflow to be in line with existing services.

This do not include implementation of the view.


cintamani:
Needs rebasing

PR #587, opened 2019-12-09
Externalise permission checks logic into services

cintamani:
As part of building a new flow with shared forms with the existing renewal flow, I have stumped across the issue of how to deal with running the correct security checks when the same form is applied on different transient objects.
In order to solve the problem, we are now using a service to check the class of the transient object we are processing and based on the result, another service is run, specific to the transient object and flow, that will run the correct checks.


app/services/waste_carriers_engine/flow_permission_checks_service.rb
```diff
@@ -0,0 +1,33 @@
...
+    private
+
+    def run_flow_setup_checks
+      params = { transient_registration: transient_registration, user: user }
+
+      case transient_registration
+      when RenewingRegistration
+        RenewingRegistrationPermissionChecksService.run(params)
+      when OrderCopyCardsRegistration
+        OrderCopyCardsRegistrationPermissionChecksService.run(params)
```
cintamani: This is the main point we should consider documenting. From now on, every new transient registration using the FormController will need to also have a rule for it in this file and a Service created for its permission checks

cintamani:
Codeclimate is a won't fix

cintamani:
@irisfaraway I will need you to hit merge as well! Codeclimate 🤷‍♀

PR #590, opened 2019-12-11
Order copy cards completion service

cintamani:

This PR contains the implementation of the order copy cards flow completion service

### DIFF
--- a/app/services/waste_carriers_engine/order_copy_cards_completion_service.rb
+++ b/app/services/waste_carriers_engine/order_copy_cards_completion_service.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class OrderCopyCardsCompletionService < BaseService
+    include CanMergeFinanceDetails
+
+    attr_reader :transient_registration
+
+    delegate :registration, to: :transient_registration
+
+    def run(transient_registration)
+      @transient_registration = transient_registration
+
+      complete_order_copy_cards
+    end
+
+    private
+
+    def complete_order_copy_cards
+      create_past_registration
+
+      update_registration
+
+      delete_transient_registration
+
+      send_confirmation_email
+    end
+
+    def create_past_registration
+      PastRegistration.build_past_registration(registration)
+    end
+
+    def update_registration
+      merge_finance_details
+
+      registration.save!
+    end
+
+    def delete_transient_registration
+      transient_registration.delete
+    end
+
+    def send_confirmation_email
+      # TODO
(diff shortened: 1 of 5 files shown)


app/services/waste_carriers_engine/order_copy_cards_completion_service.rb
```diff
@@ -0,0 +1,50 @@
...
+    def run(transient_registration)
+      @transient_registration = transient_registration
+
+      complete_order_copy_cards
+    end
+
+    private
+
+    def complete_order_copy_cards
+      create_past_registration
```
Cruikshanks: I'd want @irisfaraway to confirm, but I don't think we need this or `create_past_registration()`. If I've read `PastRegistration. build_past_registration` correctly, in this situation it is always going to return `null` and not create anything so there is no need to call it.

Plus I don't believe we should be storing a copy of the registration in past registrations just because a copy card has been ordered.

spec/services/waste_carriers_engine/order_copy_cards_completion_service_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        )
+      end
+
+      it "completes an order" do
+        orders = double(:orders)
+        payments = double(:payments)
+        transient_order = double(:transient_order)
+        transient_payment = double(:transient_payment)
+
+        # Creates a past registration
```
Cruikshanks: The method is definitely being called, but as I said In my previous comment, I don't believe it is actually creating a past registration.

So I think this test is not right, but it doesn't matter as I don't think we should be doing this anyway!

cintamani: Good! I have copied it over from the renewal completion service, I have no strong feelings on keeping it so happy to remove if Iris say so.

Cruikshanks: Let's just go ahead and remove it. It was more my reading of `build_past_registration()` I was after a second opinion on (and that in this scenario it would always return null and not create anything).

However, I'm happy to state we should not be creating a past registration record when just taking a copy card order. So go ahead and remove it from the service and the spec and then I'm good to approve.

app/services/waste_carriers_engine/order_copy_cards_completion_service.rb
irisfaraway: Agreed, this only exists if the registration is being modified. All we're doing is creating new orders and payments, which is dealt with differently than the other data.

PR #600, opened 2019-12-12
Fix back buttons on copy cards journey pages

cintamani:

Fix back buttons to send the correct token rather than a reg_identifier


cintamani:
No more likely I have copy-pasted from different places. We don't really have a convention and it doesn't make a difference for the code :) happy to update it

irisfaraway:
@cintamani I don't really mind, would rather go through and tidy en masse later if we have to. Merge away!

PR #609, opened 2019-12-17
Change validation message to represent real maximum

cintamani:

This fixes the validation message we display to the user to contain the correct maximum number of cards they can order.

### DIFF
--- a/config/locales/forms/copy_cards_forms/en.yml
+++ b/config/locales/forms/copy_cards_forms/en.yml
@@ -16,5 +16,5 @@ en:
           attributes:
             temp_cards:
               not_a_number: "You must enter a number between 1 and 999"
-              greater_than_or_equal_to: "You must enter a number between 1 and 99"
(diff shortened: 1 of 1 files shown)


config/locales/forms/copy_cards_forms/en.yml
```diff
@@ -15,6 +15,6 @@ en:
         waste_carriers_engine/copy_cards_form:
           attributes:
             temp_cards:
-              not_a_number: "You must enter a number between 1 and 99"
+              not_a_number: "You must enter a number between 1 and 999"
               greater_than_or_equal_to: "You must enter a number between 1 and 99"
```
irisfaraway:
```suggestion
              greater_than_or_equal_to: "You must enter a number between 1 and 999"
```

Missed a spot :)

cintamani: I do need 🤓

PR #616, opened 2019-12-19
Fix errors in pluralisation and remove dots

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-814

QA have found out some inconsistencies with the use of dots and pluralization of order descriptions. This PR includes fixes for both. Instructions will be give to deal with legacy data during testing.
There is also a fix about the email address used in the wireframe, which didn't match what is currently sent out to the user.

### DIFF
--- a/config/locales/mailers/order_copy_cards_mailer.en.yml
+++ b/config/locales/mailers/order_copy_cards_mailer.en.yml
@@ -4,26 +4,26 @@ en:
       send_order_completed_email:
         subject: "We’re printing your waste carriers registration cards"
         dear: "Dear %{full_name}"
-        heading: "We’re printing your waste carriers registration cards. They should arrive within 10 working days."
+        heading: "We’re printing your waste carriers registration cards. They should arrive within 10 working days"
         receipt:
           heading: "Receipt"
           order: "Order: %{description}"
           ordered_on: "Ordered on: %{formatted_time}"
-          reference: "Carrier registration number: %{reg_identifier}."
-          paid: "Paid: £%{total_paid} by debit or credit card."
+          reference: "Carrier registration number: %{reg_identifier}"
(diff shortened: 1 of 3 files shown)


spec/models/waste_carriers_engine/order_item_spec.rb
```diff
@@ -58,6 +59,15 @@ module WasteCarriersEngine
...
       end
+
+      context "when the number of cards is 1" do
+        let(:cards) { 1 }
+
+        it "should set the correct description" do
+          expect(order_item.description).to eq("1 registration card")
+        end
+      end
+
```
cintamani:
```suggestion
```

config/locales/mailers/order_copy_cards_mailer.en.yml
```diff
@@ -4,26 +4,26 @@ en:
       send_order_completed_email:
         subject: "We’re printing your waste carriers registration cards"
         dear: "Dear %{full_name}"
-        heading: "We’re printing your waste carriers registration cards. They should arrive within 10 working days."
+        heading: "We’re printing your waste carriers registration cards. They should arrive within 10 working days"
```
Cruikshanks: I have spoken with @andrewhick and confirmed this. If we are trying to ensure the pluralisation is correct for copy cards, then this should include the subject and heading in the email as well, not just the back office UI.

And as this PR is focused on fixing these issues, I thought it would make sense to make those changes here rather than as subsequent one. However, if you don't agree I'm happy to approve. We'll just need another PR for handling pluralisation in the email content.

cintamani: Since doing that requires to add a `quantity ` field to the OrderItem object, I think I will open a different PR :)

PR #621, opened 2019-12-23
Return default postcode when building XML for Worldpay

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-769

This fixes a bug on the creation of the XML sent to worldpay for which the request was rejected when a postcode was missing.

### DIFF
--- a/app/services/waste_carriers_engine/worldpay_xml_service.rb
+++ b/app/services/waste_carriers_engine/worldpay_xml_service.rb
@@ -79,7 +79,7 @@ module WasteCarriersEngine
 
       address1 = [address.house_number, address.address_line_1].join(" ")
       address2 = address.address_line_2
-      postcode = address.postcode.presence || "00000"
+      postcode = address.postcode.presence || "UNKNOWN"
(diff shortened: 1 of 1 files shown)


app/services/waste_carriers_engine/worldpay_xml_service.rb
```diff
@@ -79,7 +79,7 @@ def build_address(xml)
 
       address1 = [address.house_number, address.address_line_1].join(" ")
       address2 = address.address_line_2
-      postcode = address.postcode
+      postcode = address.postcode.presence || "00000"
```
Cruikshanks: I believe we said in the story we would set this to `"UNKNOWN"` if the postcode was missing.

Has this just been missed, or is there a reason I'm not aware of to set it to `"00000"`?

cintamani: Oh sorry! My mistake. 
```suggestion
      postcode = address.postcode.presence || "UNKNOWN"
```

PR #624, opened 2019-12-27
Ceased or revoked - start form

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-807

This adds all the code that will deal with the start of the journey for the ceased and revoked form. 

Screenshot 2019-12-27 at 18 39 42
Screenshot 2019-12-27 at 18 39 35

### DIFF
--- a/app/forms/waste_carriers_engine/cease_or_revoke_form.rb
+++ b/app/forms/waste_carriers_engine/cease_or_revoke_form.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class CeaseOrRevokeForm < BaseForm
+    def self.can_navigate_flexibly?
+      false
+    end
+
+    delegate :metaData, to: :transient_registration
+    delegate :status, :revoked_reason, to: :metaData, allow_nil: true
+    delegate :contact_address, :company_name, :registration_type, :tier, to: :transient_registration
+
+    validate :validate_status
+    validate :validate_revoked_reason
+
+    private
+
+    def validate_status
+      return true if %w[INACTIVE REVOKED].include?(status)
+
+      errors.add(:status, :presence)
+
+      false
+    end
+
+    def validate_revoked_reason
+      if revoked_reason.blank?
+        errors.add(:revoked_reason, :presence)
+
+        false
+      elsif revoked_reason.size > 500
+        errors.add(:revoked_reason, :length)
+
+        false
+      end
+
+      true
+    end
+  end
+end
--- a/config/locales/forms/cease_or_revoke_forms/en.yml
+++ b/config/locales/forms/cease_or_revoke_forms/en.yml
@@ -17,9 +17,8 @@ en:
           legend: "What do you want to do?"
           options:
             revoke: "Revoke"
-            cease: "De-register"
+            cease: "Cease"
           reason: "Reason"
-        total_cost_info: "The total cost will be shown on the next page"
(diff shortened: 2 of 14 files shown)


app/forms/waste_carriers_engine/cease_or_revoke_form.rb
```diff
@@ -0,0 +1,40 @@
...
+    delegate :status, :revoked_reason, to: :metaData, allow_nil: true
+    delegate :contact_address, :company_name, :registration_type, :tier, to: :transient_registration
+
+    validate :validate_status
+    validate :validate_revoked_reason
+
+    private
+
+    def validate_status
+      return true if %w[INACTIVE REVOKED].include?(status)
```
Cruikshanks: Just double-checking this. For example, I note `EXPIRED` is not listed.

As only an `ACTIVE` registration can be ceased or revoked, would it be easier to just check that `status == "ACTIVE"`?

config/locales/forms/cease_or_revoke_forms/en.yml
```diff
@@ -0,0 +1,35 @@
...
+          broker_dealer: "Broker and dealer"
+          carrier_broker_dealer: "Carrier, broker and dealer"
+        reason:
+          label: "Reason"
+          help_text: "Maximum 500 characters"
+        cease_or_revoke:
+          legend: "What do you want to do?"
+          options:
+            revoke: "Revoke"
+            cease: "De-register"
```
Cruikshanks: I noted it in the screenshot attached to the PR. All the error messaging refers to "cease" yet we have labelled the radio button "De-register". I know that's what it is currently called, but we're aiming to bring the terminology inline with Waste Exemptions.

```suggestion
            cease: "Cease"
```

config/locales/forms/cease_or_revoke_forms/en.yml
```diff
@@ -0,0 +1,35 @@
...
+          reason: "Reason"
+        total_cost_info: "The total cost will be shown on the next page"
```
Cruikshanks: Not 100% sure we are using this in this context 😁 !

app/forms/waste_carriers_engine/cease_or_revoke_form.rb
cintamani: 🤔 This validation is only for the form's params. The validation on the status to be active in order for the flow to work is in the permissions check. Why should this deal with expired?

config/locales/forms/cease_or_revoke_forms/en.yml
cintamani: Oh!! I thought from our last conversation that we need to keep `de-register` instead of `cease` for now

config/locales/forms/cease_or_revoke_forms/en.yml
cintamani: 😮 nope! We are not!! XD

app/forms/waste_carriers_engine/cease_or_revoke_form.rb
Cruikshanks: 🤦‍♂

config/locales/forms/cease_or_revoke_forms/en.yml
Cruikshanks: My bad. Just meant behind the scenes everything should work the same. But in the UI it's "cease" all the way.

PR #631, opened 2020-01-03
Add payment types methods and default currency

cintamani:
This creates a concern for payment types and create methods for every type.
It also adds a default value for the currency to be "GBP".


spec/models/waste_carriers_engine/payment_spec.rb
```diff
@@ -7,6 +7,18 @@ module WasteCarriersEngine
     let(:transient_registration) { build(:renewing_registration, :has_required_data) }
     let(:current_user) { build(:user) }
 
+    it_should_behave_like "Can have payment type", resource: described_class.new
+
+    describe "default attributes" do
+      describe ".currency" do
+        it "initialize a currencty to GBP" do
```
Cruikshanks:
```suggestion
        it "initialize currency as GBP" do
```

PR #637, opened 2020-01-10
[TBD] Remove token and instead use Registration's _id

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-848

This replace all instances of `token` so that now we rely on the Registration's _id instead.


cintamani:
We decided to stick with the token as the ID is not randomly generaterd

PR #642, opened 2020-01-16
Add excess amount method to finance details

cintamani:
Since there are a lot of finance operations that deal with excess amounts, such as write offs and refund, I have decided that is best to add the method that calculate the excess amount in the model so that we don't need to duplicate it in multiple helper or presenters.

### DIFF
--- a/app/models/waste_carriers_engine/finance_details.rb
+++ b/app/models/waste_carriers_engine/finance_details.rb
@@ -27,11 +27,11 @@ module WasteCarriersEngine
     end
 
     # This amounts are used in a number of finance operations.
-    def amount_in_excess
+    def overpaid_balance
       @_amount_in_excess ||= balance <= 0 ? (balance * -1) : 0
     end
 
-    def amount_to_pay
+    def unpaid_balance
       @_amount_to_pay ||= [0, balance].max
     end
 
--- a/app/models/waste_carriers_engine/finance_details.rb
+++ b/app/models/waste_carriers_engine/finance_details.rb
@@ -26,6 +26,11 @@ module WasteCarriersEngine
       finance_details
     end
 
+    # This amount is used in a large number of finance operations.
+    def amount_in_excess
+      @_amount_in_excess ||= balance <= 0 ? (balance * -1) : 0
+    end
+
     def update_balance
       order_balance = orders.sum { |item| item[:total_amount] }
       # Select payments where the type is not WORLDPAY, or if it is, the status is AUTHORISED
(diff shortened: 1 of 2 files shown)


app/models/waste_carriers_engine/finance_details.rb
```diff
@@ -26,6 +26,15 @@ def self.new_renewal_finance_details(transient_registration, method, current_use
       finance_details
     end
 
+    # This amounts are used in a number of finance operations.
+    def amount_in_excess
```
Cruikshanks: My feedback is that I think the names of these methods should be more about what they are reporting and less about the reason we have added them. Spoke with @irisfaraway because I'm useless at naming things, and she has what I think are great suggestions.

```suggestion
    def overpaid_balance
```

app/models/waste_carriers_engine/finance_details.rb
```diff
@@ -26,6 +26,15 @@ def self.new_renewal_finance_details(transient_registration, method, current_use
...
+      @_amount_in_excess ||= balance <= 0 ? (balance * -1) : 0
+    end
+
+    def amount_to_pay
```
Cruikshanks: Follows on from my previous comment

```suggestion
    def unpaid_balance
```

app/models/waste_carriers_engine/finance_details.rb
```diff
@@ -26,6 +26,15 @@ def self.new_renewal_finance_details(transient_registration, method, current_use
       finance_details
     end
 
+    # This amounts are used in a number of finance operations.
```
cintamani:
```suggestion
    # These amounts are used in a number of finance operations.
```

app/models/waste_carriers_engine/finance_details.rb
```diff
@@ -26,6 +26,15 @@ def self.new_renewal_finance_details(transient_registration, method, current_use
       finance_details
     end
 
+    # These amounts are used in a number of finance operations.
+    def overpaid_balance
+      @_amount_in_excess ||= balance <= 0 ? (balance * -1) : 0
```
cintamani:
```suggestion
      @_ overpaid_balance ||= balance <= 0 ? (balance * -1) : 0
```

app/models/waste_carriers_engine/finance_details.rb
```diff
@@ -26,6 +26,15 @@ def self.new_renewal_finance_details(transient_registration, method, current_use
...
+    end
+
+    def unpaid_balance
+      @_amount_to_pay ||= [0, balance].max
```
cintamani:
```suggestion
      @_unpaid_balance ||= [0, balance].max
```

PR #647, opened 2020-01-24
Fix broken 'Finished' link on renewal end pages

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-859

We spotted that the 'Finished' button on the 'Renewal complete' and 'Renewal received' buttons was not behaving properly. For front office users, we now want to redirect straight to the front office dashboard; for back office users, we want to redirect to the new 'view details' page in the back office.

So we'll just provide a new default method in the engine and let the host apps override it as needed.


irisfaraway:
Once this is approved and merged, I will update the Gemfiles in the host app updates (https://github.com/DEFRA/waste-carriers-back-office/pull/619 and https://github.com/DEFRA/waste-carriers-front-office/pull/381) and then this will be fixed.

PR #656, opened 2020-02-04
Add charge adjust order item type and method

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-870

This adds a new order item type of CHARGE_ADJUST.

### DIFF
--- a/app/models/waste_carriers_engine/order_item.rb
+++ b/app/models/waste_carriers_engine/order_item.rb
@@ -15,7 +15,7 @@ module WasteCarriersEngine
 
     field :amount,                          type: Integer
     field :quantity,                        type: Integer
-    field :currency,                        type: String,   default: "GBP"
+    field :currency,                        type: String, default: "GBP"
     field :lastUpdated, as: :last_updated,  type: DateTime
     field :description,                     type: String
     field :reference,                       type: String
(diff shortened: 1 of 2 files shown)


app/models/waste_carriers_engine/order_item.rb
```diff
@@ -9,12 +9,13 @@ class OrderItem
...
       edit: "EDIT",
-      copy_cards: "COPY_CARDS"
+      copy_cards: "COPY_CARDS",
+      charge_adjust: "CHARGE_ADJUST"
     )
 
     field :amount,                          type: Integer
     field :quantity,                        type: Integer
-    field :currency,                        type: String
+    field :currency,                        type: String,   default: "GBP"
```
cintamani:
```suggestion
    field :currency,                        type: String, default: "GBP"
```

spec/models/waste_carriers_engine/order_item_spec.rb
```diff
@@ -12,6 +12,14 @@ module WasteCarriersEngine
 
     let(:transient_registration) { build(:renewing_registration, :has_required_data) }
 
+    describle ".new_charge_adjust_item"
```
cintamani:
```suggestion
    describe ".new_charge_adjust_item" do
```

PR #660, opened 2020-02-07
Add granular error logging to Completion service

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-885

In order to have more data to debug the stuck renewals issue, we have decided to add more granular logging around our completion service.


cintamani:
Codeclimate is won't fix

app/services/waste_carriers_engine/renewal_completion_service.rb
```diff
@@ -36,8 +41,24 @@ def complete_renewal
...
+      raise(WrongWorkflowState, message) unless transient_registration.renewal_application_submitted?
+      raise(StillUnpaidBalance, message) if transient_registration.unpaid_balance?
+      raise(PendingConvictionCheck, message) if transient_registration.pending_manual_conviction_check?
+
+      unless %w[ACTIVE EXPIRED].include?(registration.metaData.status)
+        message = "#{message}. Status: #{registration.metaData.status}"
+        raise(WrongStatus, message)
+      end
+
+      raise(CannotComplete, message)
```
cintamani: There is no test for this, as this is "In case is not something else". Hence, if raised, we have to start from square 0 and search for a different cause.

PR #662, opened 2020-02-07
Create a new EditRegistration

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-827

When a user hits a certain route, it should start an edit and create a new EditRegistration.

This should include a permissions check, as we only want back office agency users to be able to access this.

### DIFF
--- a/app/controllers/waste_carriers_engine/edit_forms_controller.rb
+++ b/app/controllers/waste_carriers_engine/edit_forms_controller.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class EditFormsController < FormsController
+    def new
+      super(EditForm, "edit_form")
+    end
+
+    # Override this method as user shouldn't be able to "submit" this page
+    def create; end
+
+    private
+
+    # rubocop:disable Naming/MemoizedInstanceVariableName
+    def find_or_initialize_transient_registration(token)
+      @transient_registration ||= EditRegistration.where(reg_identifier: token).first ||
+                                  EditRegistration.where(token: token).first ||
+                                  EditRegistration.new(reg_identifier: token)
+    end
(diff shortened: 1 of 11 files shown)


irisfaraway:
Will be ready for review once #661 is merged.

app/controllers/waste_carriers_engine/edit_forms_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class EditFormsController < FormsController
+    def new
+      super(EditForm, "edit_form")
+    end
+
+    # Override this method as user shouldn't be able to "submit" this page
+    def create; end
```
cintamani: I know we have this pattern existing already, but double questioning it now. Shouldn't we just not create the route?

app/forms/waste_carriers_engine/edit_form.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class EditForm < BaseForm
+    delegate :reg_identifier, to: :transient_registration
```
cintamani: this is already defined in the BaseForm :) 
```suggestion
```

app/controllers/waste_carriers_engine/edit_forms_controller.rb
irisfaraway: It's only temporary for this PR as we haven't built the rest of the journey yet. So maybe something to look into for future but I don't think it's a blocker here.

app/forms/waste_carriers_engine/edit_form.rb
irisfaraway: Oops! Fixed: 90eb916

PR #666, opened 2020-02-10
Set up edit overview page

irisfaraway:

We need a new 'edit overview', as per wireframe, page that acts as the core of the edit journey. This page shows information about the registration and has links to change information.


irisfaraway:
Think this is ready for review. There is still a TODO on the date:

image

But that's because we don't actually store it anywhere yet, so I'd like to sort that out separately!

app/forms/concerns/waste_carriers_engine/can_display_data_overview.rb
```diff
@@ -0,0 +1,26 @@
...
+      delegate :location, to: :transient_registration
+      delegate :main_people, to: :transient_registration
+      delegate :phone_number, to: :transient_registration
+      delegate :registered_address, to: :transient_registration
+      delegate :registration_type, to: :transient_registration
+      delegate :relevant_people, to: :transient_registration
+      delegate :tier, to: :transient_registration
+    end
+  end
+end
```
cintamani: Since we are refactoring this, isn't it worth to have a presenter? From the name of this concerns seems to be something needed in a view only

irisfaraway: I think you are probably right! I'll change it.

irisfaraway:
@cintamani It would be but MongoDb doesn't track that by default ☹️ So I've got https://github.com/DEFRA/waste-carriers-engine/pull/668 in the works for that!

cintamani:
Ahhhh right we got MongoDB here 🤦‍♂

PR #672, opened 2020-02-13
Clean up edit view for lower tier regs

irisfaraway:

Lower tier registrations don't have some of the attributes that upper tier ones do. So this PR hides those missing values to make the edit screen look cleaner, and hide values that shouldn't be editable.

### DIFF
--- a/app/presenters/waste_carriers_engine/edit_form_presenter.rb
+++ b/app/presenters/waste_carriers_engine/edit_form_presenter.rb
@@ -62,6 +62,8 @@ module WasteCarriersEngine
     end
 
     def registration_type
+      return if transient_registration.registration_type.blank?
+
       I18n.t("#{LOCALES_KEY}.registration_type.#{transient_registration.registration_type}")
     end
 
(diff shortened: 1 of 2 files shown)


app/presenters/waste_carriers_engine/edit_form_presenter.rb
```diff
@@ -62,6 +62,8 @@ def registered_address
     end
 
     def registration_type
+      return if transient_registration.registration_type.blank?
```
cintamani: In the certificate, for lower tier, we always show `carrier, brokers or dealer` type. Shall we do the same here and default to that?

irisfaraway: Interesting, I didn't realise that about the certificate! I'm not sure, to be honest – it's not technically a value that has been set or can be modified, so I'm not sure there is much point in displaying it, especially if it's only for the eyes of back office users.

Cruikshanks: Leave it off. The user never submitted the value and its not something that we would allow to be edited for a lower-tier registration.

So I believe its simpler, clearer and more 'correct' to just not show this at all for lower-tier.

PR #673, opened 2020-02-13
Set up edit workflow

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-830

Editing is a less linear workflow than renewals or new registrations - we essentially want users to go from the overall edit page, to an existing form for an individual piece of data (eg contact phone) and then back to the overall edit page.

We should imitate the WEX implementation and behaviour.


irisfaraway:
This PR looks super long, but it is also very repetitive... so hopefully not too much of a nightmare? Let me know if you'd rather I split it up!

app/models/concerns/waste_carriers_engine/can_use_edit_registration_workflow.rb
```diff
@@ -27,10 +167,102 @@ module CanUseEditRegistrationWorkflow
...
+    private
+
+    def based_overseas?
+      overseas?
+    end
+
+    def registered_address_was_manually_entered?
+      return unless registered_address
+
+      registered_address.manually_entered?
```
cintamani: Seems like this is missing test coverage :)

app/models/concerns/waste_carriers_engine/can_use_edit_registration_workflow.rb
```diff
@@ -27,10 +167,102 @@ module CanUseEditRegistrationWorkflow
...
+    end
+
+    def skip_to_manual_address?
+      temp_os_places_error
+    end
+
+    def contact_address_was_manually_entered?
+      return unless contact_address
+
+      contact_address.manually_entered?
```
cintamani: Seems like this is missing test coverage :)

irisfaraway:
@cintamani 😲 Good spot, I definitely need to sort those out! Will fix.

app/models/concerns/waste_carriers_engine/can_use_edit_registration_workflow.rb
irisfaraway: I think this was a copy-paste goof - we don't actually use them for anything! So I've ripped 'em out.

PR #674, opened 2020-02-13
Fix missing text values on edit page

irisfaraway:

I tested this out with some alternate business types and ran into problems. Some of these are down to dodgy data (charities shouldn't really have main people, for example) but we also needed to take into account some of the business types that are still generated by the frontend. Basically, we want to fail a little more gracefully!

Also swapped a `n/a` for `-` when a location is not provided, to be consistent with the registration details page.

### DIFF
--- a/app/presenters/waste_carriers_engine/edit_form_presenter.rb
+++ b/app/presenters/waste_carriers_engine/edit_form_presenter.rb
@@ -74,8 +74,12 @@ module WasteCarriersEngine
     private
 
     def format_main_person(person)
-      role = I18n.t("#{LOCALES_KEY}.main_people.#{transient_registration.business_type}")
-      "#{person_name(person)} (#{role})"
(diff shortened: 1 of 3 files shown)


app/presenters/waste_carriers_engine/edit_form_presenter.rb
```diff
@@ -74,8 +74,12 @@ def tier
     private
 
     def format_main_person(person)
-      role = I18n.t("#{LOCALES_KEY}.main_people.#{transient_registration.business_type}")
-      "#{person_name(person)} (#{role})"
+      role = I18n.t("#{LOCALES_KEY}.main_people.#{transient_registration.business_type}", default: "")
+      if role.present?
+        "#{person_name(person)} (#{role})"
+      else
+        person_name(person)
```
cintamani: Seems like codeclimate is complaining about missing test coverage for this `else`, shall we add some?

irisfaraway: Added an extra test to cover this 👍

PR #678, opened 2020-02-14
Improve certificates for lower tier sole traders and partnerships

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-834

The CertificatePresenter tries to make more use of the "main people" names when the business type is sole trader or partnership. However, for lower tier regs, we never ask for the main people. This meant the certificate sometimes displayed blank values where a person's name would be.

So this change only applies that extra logic for upper tier partnerships and sole traders.

### DIFF
--- a/app/presenters/waste_carriers_engine/certificate_presenter.rb
+++ b/app/presenters/waste_carriers_engine/certificate_presenter.rb
@@ -6,7 +6,7 @@ module WasteCarriersEngine
     LOCALES_KEY = ".waste_carriers_engine.pdfs.certificate"
 
     def carrier_name
-      return company_name unless business_type == "soleTrader"
+      return company_name unless upper_tier? && business_type == "soleTrader"
 
       # Based on the logic found in the existing certificate, we simply display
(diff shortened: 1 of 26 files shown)


app/presenters/waste_carriers_engine/certificate_presenter.rb
```diff
@@ -6,7 +6,7 @@ class CertificatePresenter < BasePresenter
     LOCALES_KEY = ".waste_carriers_engine.pdfs.certificate"
 
     def carrier_name
-      return company_name unless business_type == "soleTrader"
+      return company_name unless upper_tier? && business_type == "soleTrader"
```
cintamani: I do find reading an `unless` with `&&` very very hard!! Can we make it an `if` and/or make a method for the check only? 🙇‍♀

irisfaraway: Agreed, it is messy! I've tidied it up 👍

cintamani: And I love it now! 😍

PR #691, opened 2020-02-24
Deal with potential location changes in overseas? check

irisfaraway:
There was a gap in our logic for https://github.com/DEFRA/waste-carriers-engine/pull/687. If a user had previously said they were based overseas when registering, `business_type` was set to `overseas`. Then if they selected a different location on renewal, `overseas?` would still respond true.

So this PR adds an extra `uk_location?` check to make sure we don't have this happen!

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
+++ b/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
@@ -95,7 +95,13 @@ module WasteCarriersEngine
         business_type == "charity"
       end
 
+      def uk_location?
+        %w[england scotland wales northern_ireland].include?(location)
+      end
+
       def overseas?
+        return false if uk_location?
+
         location == "overseas" || business_type == "overseas"
       end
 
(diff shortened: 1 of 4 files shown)


irisfaraway:
Now failing due to expired cassettes 🙄 Should be fixed once https://github.com/DEFRA/waste-carriers-engine/pull/690 is merged in.

irisfaraway:
@Cruikshanks Fixed the test!

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -95,7 +95,13 @@ def charity?
         business_type == "charity"
       end
 
+      def uk_location?
+        %w[england scotland wales northern_ireland].include?(location)
```
cintamani: Shall we make those constants in some relevant place?

irisfaraway: I definitely think we should do this – however I started working on it and it grew in scope pretty rapidly! So I may tackle it in a separate PR if that is OK with you?

PR #693, opened 2020-02-25
Add new transient registration type for NewRegistration

cintamani:
This is milestone 1 for our work on New registrations. This simply add a new model, along with an empty state machine, a basic test and an empty factory.


spec/models/waste_carriers_engine/new_registration_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module WasteCarriersEngine
+  RSpec.describe NewRegistration, type: :model do
+    subject(:renewing_registration) { build(:new_registration) }
```
cintamani:
```suggestion
    subject(:new_registration) { build(:new_registration) }
```

PR #694, opened 2020-02-26
Rename shared post form example to specify renewal specific

cintamani:
This renames the shared "POST form" examples to "POST renewal form". This is because the current implementation of the shared examples is very specific to renewals in the fact that it needs authentication and the object used is a renewal registration.


cintamani:
Actually realised we don't need to do this. We will have to instead remove the check for a logged in used across all those forms as they will need to be shared with the NewRegistration flow. Authentication checks will stay only on start forms.

cintamani:
Actually no, I re-changed my mind. I think this change is OK as this has also tests on the `renewable` part of posting a renewal form. Hence for now this will `free` the `POST form` name so we can develop a more generic one and then refactor this to only contain `renewal` specific scenarios

PR #695, opened 2020-02-26
Add start form for new registrations

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-840

This adds a new page to the services which allow a user to start a new registration journey.
The page is part of the new registration workflow.

I suggest reviewing by separate commits

Screenshot 2020-02-26 at 15 22 42
Screenshot 2020-02-26 at 15 14 45


Cruikshanks:
@cintamani I've removed myself and @irisfaraway as reviewers as I know we suggested the tweak about prefixing the start option with `temp` plus assuming once the other PR's are merged this will undergo a bit of refactoring.

So figured it makes our TODO list a little clearer if we remove ourselves for now and then you add us again when the changes have been made.

cintamani:
Yes definitely! Thanks guys :) will ping you when it is ready again

app/controllers/waste_carriers_engine/renew_registration_forms_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    def create
+      super(RenewRegistrationForm, "renew_registration_form")
+    end
+
+    private
+
+    def transient_registration_attributes
+      params.fetch(:renew_registration_form, {}).permit()
```
cintamani:
```suggestion
      params.fetch(:renew_registration_form, {})
```

cintamani:
The missing test coverage is for the temproary renewing registration forms and controller which will be implemented and properly tested as per RUBY-841

app/views/waste_carriers_engine/start_forms/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        <legend class="visuallyhidden">
+          <%= t(".heading") %>
+        </legend>
+
+        <% if @start_form.errors[:start_option].any? %>
+          <span class="error-message"><%= @start_form.errors[:start_option].join(", ") %></span>
+        <% end %>
+
+        <div class="multiple-choice">
+          <%= f.radio_button :start_option, 'new' %>
```
irisfaraway: Can we use the START_OPTIONS constant at all here? :bowtie:

spec/forms/waste_carriers_engine/base_forms_spec.rb
```diff
@@ -46,6 +46,7 @@ module WasteCarriersEngine
         context "when the resource is not a transient_registration" do
```
irisfaraway: Is this accurate? A NewRegistration is still a type of TransientRegistration.

spec/models/waste_carriers_engine/new_registration_workflow/start_form_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module WasteCarriersEngine
+  RSpec.describe NewRegistration do
+    subject(:new_registration) { build(:new_registration, start_option: start_option) }
+
+    describe "#workflow_state" do
+      context ":start_form state transitions" do
```
irisfaraway: Do we need a test that this cannot go `back`?

spec/requests/waste_carriers_engine/start_forms_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+          post new_start_form_path(params)
+
+          new_registration = WasteCarriersEngine::NewRegistration.last
+
+          expect(new_registration).to be_present
+        end
+      end
+
+      context "when a new registration token is not passed to the request and params are invalid" do
+        it "does creates a new transient registration of type NewRegistration" do
```
irisfaraway:
```suggestion
        it "does not creates a new transient registration of type NewRegistration" do
```

app/views/waste_carriers_engine/start_forms/new.html.erb
cintamani: Oh yes! I forgot to make that fix! Thanks for pointing that out :D

spec/forms/waste_carriers_engine/base_forms_spec.rb
cintamani: yes, this is not testing the validity of a `new_regisration` object in the base_form, but the validity of a `registration` object passed as a transient registration, so is still correct. You could argue that there is no test coverage for when we have a transient registration that is of type new registration, but as we discussed during the dev meeting we aren't sure about keeping this validation at all, and I will have to sort it out for validating the correct CBDU/CBDL numbers are generated later on so I just added that `is_a?` as a placeholder and make things pass now. We will need to go in more details about that in some other PR/meeting :D

spec/models/waste_carriers_engine/new_registration_workflow/start_form_spec.rb
cintamani: I lean towards not testing that something isn't defined in cases like this. After all, the `next` and `back` event is not something built-in that we get by default, but something we do deliberately define. Hence, I don't think we need that test coverage.

app/views/waste_carriers_engine/start_forms/new.html.erb
cintamani:
```suggestion
          <%= f.radio_button :start_option, WasteCarriersEngine::StartForm::NEW %>
```

app/views/waste_carriers_engine/start_forms/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+          <%= f.label :start_option, t(".options.new"), value: 'new' %>
```
cintamani:
```suggestion
          <%= f.label :start_option, t(".options.new"), value: WasteCarriersEngine::StartForm::NEW %>
```

app/views/waste_carriers_engine/start_forms/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        </div>
+        <div class="multiple-choice">
+          <%= f.radio_button :start_option, 'renew' %>
```
cintamani:
```suggestion
          <%= f.radio_button :start_option, WasteCarriersEngine::StartForm::RENEW %>
```

app/views/waste_carriers_engine/start_forms/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+          <%= f.label :start_option, t(".options.renew"), value: 'renew' %>
```
cintamani:
```suggestion
          <%= f.label :start_option, t(".options.renew"), value: WasteCarriersEngine::StartForm::RENEW %>
```

spec/forms/waste_carriers_engine/base_forms_spec.rb
irisfaraway: So the first `allow` is making sure that `is_a?(TransientRegistration)` always returns false, but the second `allow` makes sure we still actually call `is_a?(NewRegistration)` and use the real response – am I understanding this right?

cintamani: No, the second call is stubbing `is_a?(NewRegistration)` and making it return `nil`

app/models/waste_carriers_engine/new_registration.rb
```diff
@@ -3,5 +3,7 @@
 module WasteCarriersEngine
   class NewRegistration < TransientRegistration
     include CanUseNewRegistrationWorkflow
+
+    field :start_option, type: String
```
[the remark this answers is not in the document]
cintamani: Oh yeah! I did forget about that too 🤦‍♂ Will make the change :D

PR #706, opened 2020-03-05
Set Lower Tier and Upper tier in new registration journey

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-836

This extend the new registration flow to deal with the correct setting of lower tiers and upper tiers on the registration.
It includes getting rid of the placeholder lower tier screen from the renewal journey, set up the tier in the transient object as per flows and make sure the user can go back, change responses to question and change its tier accordingly. 

A different PR will deal with the rest of the journey for LT where we don't need to go through all the screens

### DIFF
--- a/spec/models/waste_carriers_engine/new_registration_workflow/company_name_form_spec.rb
+++ b/spec/models/waste_carriers_engine/new_registration_workflow/company_name_form_spec.rb
@@ -19,13 +19,31 @@ module WasteCarriersEngine
         end
 
         context "on back" do
-          context "when the business requires a registration number" do
-            subject { build(:new_registration, workflow_state: "company_name_form", business_type: "limitedCompany") }
+          context "when the business does not requires a registration number and the registration is upper tier" do
+            subject { build(:new_registration, :upper, workflow_state: "company_name_form", location: "overseas") }
 
-            include_examples "has back transition", previous_state: "registration_number_form"
+            include_examples "has back transition", previous_state: "cbd_type_form"
           end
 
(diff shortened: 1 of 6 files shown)


spec/models/waste_carriers_engine/new_registration_workflow/company_name_form_spec.rb
```diff
@@ -19,13 +19,31 @@ module WasteCarriersEngine
         end
 
         context "on back" do
-          context "when the business requires a registration number" do
-            subject { build(:new_registration, workflow_state: "company_name_form", business_type: "limitedCompany") }
+          context "when the business does not requires a registration number and the registration is upper tier" do
```
irisfaraway: Any chance we could use these as nested contexts? I think it would be easier to read, and also clearer what the "default" is (in this case, upper tier and a business type that requires a reg number, but right now none of that is explicit).

eg

```
context "when the registration is upper tier" do
  context "when the registration does not require a registration number" do
    ...
  end
  context "when the registration requires a registration number" do
    ... # I thought we were missing this one at first, but then realised it was at the bottom!
  end
end
context "when the registration is lower tier" do
  ...
end
```

cintamani: We don't have a case for `Upper tier` and `Registration number required`. I mean we do, but that's the default behaviour which we don't enforce with `if` or`unless`, hence I don't have a context for it.

cintamani:
Those upper and lower tier tickets only get up to the declaration page. We got other tickets for the final bits :) everything else just work the same way for both journeys

PR #709, opened 2020-03-06
Implement forks for lower tier new registration journey

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-836

Following on https://github.com/DEFRA/waste-carriers-engine/pull/706, this implements the forks to the journey when a registration is a lower tier rather than an upper tier.


cintamani:
Pushing a different branch or hakiry won't trigger

PR #717, opened 2020-03-10
Actually fix mongoid-locker dependency

Cruikshanks:
PR #716 was supposed to fix this. It would if I had referenced the gem properly.

Because of the way I referenced the gem previously 'only' the `mongoid/locker` code was being required rather than the whole gem 🤦

This seemed to work, but actually when you ran the back office tests you got errors because `mongoid/locker/wrapper` being referenced.

THIS will fix things 😰


Cruikshanks:
shame

PR #720, opened 2020-03-10
Generate unique identifier number on declaration page

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-925

We have decided that we will generate the unique identifier number when a declaration is signed. We will then apply a `CBDU` or `CBDL` prefix once the application is submitted


cintamani:
- Leave counter with next number ready for use 
- Check https://github.com/mongodb/mongoid/blob/196e969840c3bd35dd7517c898b0a29168d6021e/lib/mongoid/contextual/mongo.rb#L188
- check https://github.com/DEFRA/waste-carriers-service/blob/master/src/main/java/uk/gov/ea/wastecarrier/services/resources/RegistrationsResource.java#L172

cintamani:
Mostly thanks to you @Cruikshanks! :D

PR #733, opened 2020-03-16
Set up order models for type change during edit

irisfaraway:

This PR covers setting up the order models to be usable with the edit journey, rather than assume they are only used for renewals.

### DIFF
--- a/app/models/waste_carriers_engine/order.rb
+++ b/app/models/waste_carriers_engine/order.rb
@@ -90,7 +90,11 @@ module WasteCarriersEngine
     end
 
     def generate_description
-      self.description = order_items.map(&:description).join(", plus ")
+      self.description = order_items.map(&:description)
+                                    .join(", plus ")
+
+      # Upcase the first letter. (TODO: Use .upcase_first when we switch to Rails 5)
+      description[0] = description[0].capitalize
     end
 
     def update_after_worldpay(status)
--- a/app/models/waste_carriers_engine/order.rb
+++ b/app/models/waste_carriers_engine/order.rb
@@ -38,7 +38,7 @@ module WasteCarriersEngine
       # TODO: Review whether card_count.present? is still necessary - this was a fix put in to deal with WC-498
       order[:order_items] << OrderItem.new_copy_cards_item(card_count) if card_count.present? && card_count.positive?
 
-      order.generate_description
+      order.set_description
 
       order[:total_amount] = order[:order_items].sum { |item| item[:amount] }
 
@@ -89,18 +89,25 @@ module WasteCarriersEngine
       Time.now.to_i.to_s
     end
 
-    def generate_description
-      self.description = order_items.map(&:description)
-                                    .join(", plus ")
-
-      # Upcase the first letter. (TODO: Use .upcase_first when we switch to Rails 5)
-      description[0] = description[0].capitalize
+    def set_description
+      self.description = generate_description
     end
 
     def update_after_worldpay(status)
       self.world_pay_status = status
       self.date_last_updated = Time.current
       save!
+    end
+
+    private
+
+    def generate_description
+      description = order_items.map(&:description)
+                               .join(", plus ")
+
+      description[0] = description[0].capitalize
+
+      description
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/models/waste_carriers_engine/order.rb
```diff
@@ -88,7 +90,11 @@ def generate_id
     end
 
     def generate_description
-      self.description = order_items.map(&:description).join(", plus ")
+      self.description = order_items.map(&:description)
+                                    .join(", plus ")
+
+      # Upcase the first letter. (TODO: Use .upcase_first when we switch to Rails 5)
+      description[0] = description[0].capitalize
```
cintamani: Why using `self.` above but not below?

spec/models/waste_carriers_engine/order_item_spec.rb
```diff
@@ -36,12 +36,13 @@ module WasteCarriersEngine
       end
 
       it "should set the correct description" do
-        expect(order_item.description).to eq("Renewal of registration")
+        expect(order_item.description).to eq("renewal of registration")
```
cintamani: I am puzzled by this. Given that we have https://github.com/DEFRA/waste-carriers-engine/pull/733/files#diff-1aaae17839a245710574a863973dae8bR97, shouldn't this also be setting a description that starts with a capital letter?

spec/models/waste_carriers_engine/order_item_spec.rb
```diff
@@ -55,8 +56,20 @@ module WasteCarriersEngine
         expect(order_item.amount).to eq(2_500)
       end
 
-      it "should set the correct description" do
-        expect(order_item.description).to eq("changing carrier type during renewal")
+      context "when the journey is :renewal" do
+        let(:journey) { :renewal }
+
+        it "should set the correct description" do
+          expect(order_item.description).to eq("changing carrier type during renewal")
```
cintamani: I think we discussed about the possibility to have a unique description which will apply to both journeys so that we don't have to make these special cases, was that ruled out then?

spec/models/waste_carriers_engine/order_item_spec.rb
irisfaraway: I moved the capitalising to the Order model as otherwise we ended up with some wonky capitalising in the joined-up description there. I figured it was better to make the individual OrderItems consistent and then capitalise them when we combine them.

spec/models/waste_carriers_engine/order_item_spec.rb
irisfaraway: I'm not sure to be honest - I don't remember talking about this.

app/models/waste_carriers_engine/order.rb
irisfaraway: Honestly I'm not sure - it looks weird to me as well! Rubocop complains if the second one is `self.description` and says it is redundant, but the first doesn't work without it 🤷‍♀

spec/models/waste_carriers_engine/order_item_spec.rb
cintamani: Ohhh I got it now! Make sense! :D

spec/models/waste_carriers_engine/order_item_spec.rb
cintamani: I think that both Andrew and Emi would be good if we change this to a simple `changing carrier type` so let's ask them tomorrow :D

app/models/waste_carriers_engine/order.rb
cintamani: rubocop nonsense 😂 
I was thinking, what if we make one method for the setting and one method for generating the description? I think it will be clearer and rubocop should not complain! So we can have:
```
# Public interface
def set_description
  self.description = generate_description
end


...
# Private interface 
def generate_description
  description = ""
  
  # Do stuff

  description
end
```

irisfaraway: Ooh, I like this 👍 Will sort it out!

irisfaraway: Done: https://github.com/DEFRA/waste-carriers-engine/pull/733/commits/39834d7f2ccfae68f4e0ccae2c65b767fe7d19c0

PR #734, opened 2020-03-16
Add presenter for order details

irisfaraway:

This presenter will be used to display the individual charges of an order, as well as the total cost.

The aim is to standardise our various payment pages as much as we can.

### DIFF
--- a/app/presenters/waste_carriers_engine/order_and_total_presenter.rb
+++ b/app/presenters/waste_carriers_engine/order_and_total_presenter.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class OrderAndTotalPresenter < BasePresenter
+    LOCALES_KEY = ".waste_carriers_engine.shared.order_and_total"
+
+    def order_items
+      items = []
+
+      transient_registration.finance_details.orders.first.order_items.each do |item|
+        items << add_order_item(item)
+      end
+
+      items
+    end
+
+    def total_cost
+      transient_registration.finance_details.balance
+    end
+
+    private
+
+    def add_order_item(item)
+      formatted_item = {}
+
+      formatted_item[:cost] = item.amount
+      formatted_item[:description] = description_for(item)
+
+      formatted_item
+    end
+
+    def description_for(item)
+      case item.type
+      when OrderItem::TYPES[:renew]
+        description_for_renew
+      when OrderItem::TYPES[:edit]
+        description_for_type_change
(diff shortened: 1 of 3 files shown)


app/presenters/waste_carriers_engine/order_and_total_presenter.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def total_cost
+      transient_registration.finance_details.balance
+    end
+
+    private
+
+    def add_order_item(item)
+      formatted_item = {}
```
cintamani: Can we make this a Struct or create a presenter for items? Will look a bit odd otherwise that we don't have the ability to do `order_item.cost` and instead we must do `order_item[:cost]`. Also, since in the database layer we are calling this cost `amount`, I think there will be less confusion if we don't rename it and keep going with the amount to be consistent with what we usually know the cost is called on an order item

irisfaraway: Agree about `cost` vs `amount` – I've renamed it!

I'm not sure about adding a Struct or an additional presenter though – it feels kinda overkill for what I end up doing with it in the HTML (which I haven't made a separate PR for yet, but is basically just this: https://github.com/DEFRA/waste-carriers-engine/blob/2878113bd25cc3c02174b6080a2f1f01066f4d93/app/views/waste_carriers_engine/shared/_order_item.html.erb) What would you say the advantage of this would be?

cintamani: Well, the goal of creating all this is reusability :D 

Looking at the code, to me is an extra reason to not make this object a simple Hash.

This is because the partial there is not telling me that the `order_item` object IS NOT an actual `WasteCarriersEngine::OrderItem` instance (with mongoId order_item[:details] would work on the model too), and someone reading that code might assume it is. 
Hence I still think would be good if we can make sure the object there will eventually behave as an `order_item`, which is the exact reason why we use SimpleDelegator with our presenters, so we can easely extend the code using them :)

irisfaraway: That shared partial is only used within this other shared partial: https://github.com/DEFRA/waste-carriers-engine/blob/2878113bd25cc3c02174b6080a2f1f01066f4d93/app/views/waste_carriers_engine/shared/_order_and_total.html.erb

So if we're concerned about clarity we could always just merge the two as they probably don't need to be separate. I feel like it is unlikely someone will be using that individual table row on its own 🤷‍♀

PR #736, opened 2020-03-16
Make EditCompletionService deal with new finance info

irisfaraway:

The EditCompletionService currently disregards any finance_details on the transient object. However, if the user is required to pay for an edit, a new finance_details will be created. We need to be able to merge any new orders and payments onto the original registration.

The OrderCopyCardsCompletionService already deals with this, so we just need to make some tweaks the EditCompletionService to follow that example.

### DIFF
--- a/app/services/waste_carriers_engine/edit_completion_service.rb
+++ b/app/services/waste_carriers_engine/edit_completion_service.rb
@@ -2,9 +2,12 @@
 
 module WasteCarriersEngine
   class EditCompletionService < BaseService
+    attr_reader :transient_registration
+
+    delegate :registration, to: :transient_registration
+
     def run(edit_registration:)
-      @edit_registration = edit_registration
-      @registration = @edit_registration.registration
+      @transient_registration = edit_registration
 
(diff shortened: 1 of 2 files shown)


app/services/waste_carriers_engine/edit_completion_service.rb
```diff
@@ -15,39 +20,40 @@ def run(edit_registration:)
...
-                                                                 "temp_company_postcode",
-                                                                 "temp_contact_postcode",
-                                                                 "temp_os_places_error",
-                                                                 "temp_payment_method",
-                                                                 "temp_tier_check",
-                                                                 "_type",
-                                                                 "workflow_state")
-
-      @registration.write_attributes(copyable_attributes)
+      copyable_attributes = @transient_registration.attributes.except("_id",
```
cintamani:
```suggestion
      copyable_attributes = transient_registration.attributes.except("_id",
```

irisfaraway: Done ✅

PR #737, opened 2020-03-16
Build finance_details object for edit journey

irisfaraway:

This PR sets up a BuildEditFinanceDetailsService class which is used to create finance_details objects for the edit journey. It's pretty similar to OrderAdditionalCardsService so there's probably room for consolidating once we have a consistent way of building this objects (I'm looking at you, renewals).


irisfaraway:
I think tests will keep failing until https://github.com/DEFRA/waste-carriers-engine/pull/733 is merged.

PR #741, opened 2020-03-17
Set up workflow for edit payments by bank transfer

irisfaraway:

This PR sets up the workflow and pages for paying for an edit by bank transfer. This includes the payment summary and bank transfer info pages.

WorldPay will be added separately in a future PR.

### DIFF
--- a/app/controllers/waste_carriers_engine/edit_bank_transfer_forms_controller.rb
+++ b/app/controllers/waste_carriers_engine/edit_bank_transfer_forms_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class EditBankTransferFormsController < FormsController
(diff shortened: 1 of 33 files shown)


irisfaraway:
This PR looks big, but should be manageable commit-by-commit.

cintamani:
I am wondering since the layout of the pages is almost the same, why are we generating new forms for these and not reusing the existing ones we already share with new and renewing registrations?

irisfaraway:
@cintamani We already had different controllers for the copy cards payment pages, so I was following that pattern. There are some differences in the text around options, for example. Although I think that with the way I've rewritten the code, we may now be able to reuse it across all the different payment pages.

I was thinking of getting this merged in as-is and then merging all the various duplicated payment forms in a separate PR so as not to bloat this one even more – what do you think?

PR #743, opened 2020-03-18
Generate a new registration on completion with pending payments

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-916

This adds the code necessary to generate a new registration from a transient registration on the new registration journey.
It also add a first received page for upper tier journeys that end with a bank transfer pending payment

Screenshot 2020-03-18 at 19 11 01


cintamani:
Code smells from SonarCloud are a won't fix in-line with our convention

Cruikshanks:
I might have to merge this for you @cintamani . The Sonarcloud issues are false positives, but there is no way for me to mark them so whilst in the PR 😡 .

PR #747, opened 2020-03-20
Show the user a completed screen for conviction checks required 

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-917

Add the code necessary to show the user a proper completion screen for when they complete their UT paying by bank transfer but they have pending conviction checks.

Depends on: https://github.com/DEFRA/waste-carriers-engine/pull/743

Screenshot 2020-03-20 at 12 23 28

### DIFF
--- a/app/views/waste_carriers_engine/registration_received_pending_conviction_forms/new.html.erb
+++ b/app/views/waste_carriers_engine/registration_received_pending_conviction_forms/new.html.erb
@@ -15,7 +15,7 @@
 
   <p>
     <%= t(".contact.paragraph_1") %>
-    <a href="mailto:<%= t(".contact.email") %>nccc-carrierbroker@environment-agency.gov.uk"><%= t(".contact.email") %></a>
+    <a href="mailto:<%= t(".contact.email") %>"><%= t(".contact.email") %></a>
     <%= t(".contact.text") %>
   </p>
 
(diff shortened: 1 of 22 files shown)


cintamani:
Sonar and Hakiri are won't fix / ignore

cintamani:
Oh yes! I have to change the route, we shouldn't tell that to the user. 
About having one screen rather than 2 for application received, I don't think is a good idea because will make the view logic more complex and maintainability will suffer as a result.

config/routes.rb
```diff
@@ -30,6 +30,12 @@
               only: :new,
               path: "registration-received-pending-payment",
               path_names: { new: "" }
+
+    resources :registration_received_pending_conviction_forms,
+              only: :new,
+              path: "registration-received-pending-conviction",
```
cintamani:
```suggestion
              path: "registration-received",
```

irisfaraway:
Discussed this in more detail on Slack and agree to leave these controllers separate.

app/views/waste_carriers_engine/registration_received_pending_conviction_forms/new.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <h2 class="heading-medium"><%= t(".subheading_1") %></h2>
+  <p><%= t(".paragraph_2") %></p>
+
+  <div class="panel">
+    <p><%= t(".paragraph_3") %></p>
+  </div>
+
+  <p>
+    <%= t(".contact.paragraph_1") %>
+    <a href="mailto:<%= t(".contact.email") %>nccc-carrierbroker@environment-agency.gov.uk"><%= t(".contact.email") %></a>
```
irisfaraway:
```suggestion
    <a href="mailto:<%= t(".contact.email") %>"><%= t(".contact.email") %></a>
```

cintamani: Copy-paste hell! :joy:

PR #748, opened 2020-03-20
Add completed registration screen

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-918

This adds screen and status for when an upper tier registration is completed at the end of a journey

Blocked by: https://github.com/DEFRA/waste-carriers-engine/pull/743

Screenshot 2020-03-20 at 15 17 41

PR #750, opened 2020-03-23
Add completed screen for lower tier registrations

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-838

Add screens for LT completed registration.

Screenshot 2020-03-23 at 10 27 22


irisfaraway:
I think we can ignore SonarCloud on this. (Or ideally find a way to override that rule!)

cintamani:
Closing this as it will need to be built on top of: https://github.com/DEFRA/waste-carriers-engine/pull/748 as the wireframes are almost identical

PR #752, opened 2020-03-23
Only merge finance details if reg type has changed in edit

irisfaraway:

Previously we would always call `merge_finance_details` when completing an edit. This is because if the user never changed the reg type, no finance_details were created and so there was nothing to merge.

However, if the user changes the type, goes to the payment page (generating a finance_details object) and then goes back and reverts their original change, the finance_details object will still exist. This will result in an extra order on the registration that the user has no way of paying, and also shouldn't actually pay.

To avoid this happening, we add an extra check to tell the EditCompletionService to only try to merge the finance details when there is something worth keeping.


cintamani:
Cassettes have expired again

PR #756, opened 2020-03-23
Backfill individual DOB fields when initialising key person

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-830

We spotted that when editing a sole trader's key person, the birthday wasn't always pre-populated.

This turned out to be because any registrations generated in the frontend and not since modified did not store values for `dob_day`, `dob_month` and `dob_year`. Instead they just had the single `dob`. But any records generated in the new engine have all 4 values.

To get around this, we set the values of the individual fields to match the value of `dob`. This happens when a new key person is initialised, specifically when we load an older one from the database.

### DIFF
--- a/app/models/waste_carriers_engine/key_person.rb
+++ b/app/models/waste_carriers_engine/key_person.rb
@@ -10,6 +10,7 @@ module WasteCarriersEngine
 
     accepts_nested_attributes_for :conviction_search_result
 
+    after_initialize :set_individual_dob_fields
     after_initialize :set_date_of_birth
 
     field :first_name,  type: String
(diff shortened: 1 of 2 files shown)


app/models/waste_carriers_engine/key_person.rb
```diff
@@ -39,5 +40,14 @@ def set_date_of_birth
     rescue TypeError
       errors.add(:dob, :invalid_date)
     end
+
+    def set_individual_dob_fields
+      return if dob_year.present? & dob_month.present? & dob_day.present?
```
cintamani: Any particular reason why choosing the binary operator over the logical one? Just curious :D

irisfaraway: No idea, just tired I think! I'll change it

cintamani: Well, since `present?` returns true/false it would have been fine :D But yes, better to be consistent :D

irisfaraway:
Build has passed, Travis just isn't reporting back: https://travis-ci.com/github/DEFRA/waste-carriers-engine/builds/154772988

PR #761, opened 2020-03-24
Add your tier page to new registration flow

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-946

Add a new page after the smart answers to tell the user about what tier they are and why.

Screenshot 2020-03-24 at 13 03 35
Screenshot 2020-03-24 at 13 04 27


app/models/concerns/waste_carriers_engine/can_use_new_registration_workflow.rb
```diff
@@ -133,21 +138,25 @@ module CanUseNewRegistrationWorkflow
...
 
           transitions from: :waste_types_form,
-                      to: :company_name_form,
+                      to: :your_tier_form,
                       if: :switch_to_lower_tier_based_on_smart_answers?,
                       after: :switch_to_lower_tier
 
           transitions from: :waste_types_form,
-                      to: :cbd_type_form,
+                      to: :your_tier_form,
```
irisfaraway: Do we have two `from: :waste_types_form, to: :your_tier_form` transitions just because they have different `after`s? I feel like if they have the same journey then we should be setting the tier somewhere outside of the state machine, since it's not really anything to do with it. Then we can have just one transition.

spec/models/waste_carriers_engine/new_registration_workflow/your_tier_form_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+            include_examples "has back transition", previous_state: "business_type_form"
+          end
+
+          include_examples "has back transition", previous_state: "construction_demolition_form"
+        end
+      end
+    end
+  end
+end
```
irisfaraway: This must have been fun to write 😂

spec/models/waste_carriers_engine/new_registration_workflow/your_tier_form_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+            context "when the company only carries own waste" do
+              subject { build(:new_registration, :lower, workflow_state: "your_tier_form", other_businesses: "no") }
+
+              include_examples "has back transition", previous_state: "construction_demolition_form"
+            end
+
+            include_examples "has back transition", previous_state: "construction_demolition_form"
+          end
+
+          context "when the company only deals with amf waste" do
```
irisfaraway:
```suggestion
          context "when the company deals with more than amf waste" do
```

I think the options are:

"We only deal with these types of waste" - `yes`
"We deal with more than just the waste listed above" - `no`

spec/models/waste_carriers_engine/new_registration_workflow/your_tier_form_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+            include_examples "has back transition", previous_state: "waste_types_form"
+          end
+
+          context "when the registration's company is a charity" do
+            subject { build(:new_registration, workflow_state: "your_tier_form", business_type: "charity") }
+
+            include_examples "has back transition", previous_state: "business_type_form"
+          end
+
+          include_examples "has back transition", previous_state: "construction_demolition_form"
```
irisfaraway: I would like to have some context around this example, as there are so many examples above it that it can be hard to tell when this one applies at a glance.

app/models/concerns/waste_carriers_engine/can_use_new_registration_workflow.rb
cintamani: yes, is because of the different `:afters`. Currently, I feel like having the state machine handling this is the best choice, rather than having this responsibility going around in a trillion of forms or controller. I believe is more manageable and improves the maintainability. In any case, I feel like currently is too big of a refactoring/task to change the responsibility of setting the tier from the state machine to a controller, hence if that's the way we want going forward let's open a task somewhere and deal with it separately.

spec/models/waste_carriers_engine/new_registration_workflow/your_tier_form_spec.rb
cintamani: oh yes! Quite a headache dealing with the smart answers! So many rules 🙈

spec/models/waste_carriers_engine/new_registration_workflow/your_tier_form_spec.rb
cintamani: yes this is true! I changed this code a couple of times and forgot to adjust the description! Thanks! :)

spec/models/waste_carriers_engine/new_registration_workflow/your_tier_form_spec.rb
cintamani: There is no description because this is the `default` behaviour, the one without `if`

PR #767, opened 2020-03-26
Add renewal number lookup page

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-841

When external users enter the service, they see a page that asks what the user wants to do - start a new registration or renew. If they select 'renew', they see this page: 'What is your waste carrier registration number?'

We need to replace this page from the frontend with a new version.

### DIFF
--- a/app/forms/waste_carriers_engine/renew_registration_form.rb
+++ b/app/forms/waste_carriers_engine/renew_registration_form.rb
@@ -5,5 +5,11 @@ module WasteCarriersEngine
     delegate :temp_lookup_number, to: :transient_registration
 
     validates_with RenewalLookupValidator
+
+    def submit(params)
+      params[:temp_lookup_number].upcase! if params[:temp_lookup_number].present?
+
+      super
+    end
   end
 end
--- a/app/forms/waste_carriers_engine/renew_registration_form.rb
+++ b/app/forms/waste_carriers_engine/renew_registration_form.rb
@@ -2,5 +2,6 @@
 
 module WasteCarriersEngine
   class RenewRegistrationForm < BaseForm
+    delegate :temp_lookup_number, to: :transient_registration
   end
 end
--- a/app/forms/waste_carriers_engine/renew_registration_form.rb
+++ b/app/forms/waste_carriers_engine/renew_registration_form.rb
@@ -3,5 +3,7 @@
 module WasteCarriersEngine
   class RenewRegistrationForm < BaseForm
     delegate :temp_lookup_number, to: :transient_registration
+
+    validates_with RenewalLookupValidator
   end
 end
--- a/app/forms/waste_carriers_engine/renew_registration_form.rb
+++ b/app/forms/waste_carriers_engine/renew_registration_form.rb
@@ -7,7 +7,7 @@ module WasteCarriersEngine
     validates_with RenewalLookupValidator
 
     def submit(params)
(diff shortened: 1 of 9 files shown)


app/forms/waste_carriers_engine/renew_registration_form.rb
```diff
@@ -2,5 +2,14 @@
 
 module WasteCarriersEngine
   class RenewRegistrationForm < BaseForm
+    delegate :temp_lookup_number, to: :transient_registration
+
+    validates_with RenewalLookupValidator
+
+    def submit(params)
+      params[:temp_lookup_number].upcase! if params[:temp_lookup_number].present?
```
cintamani, line 10: NIT 
```suggestion
      params[:temp_lookup_number]&.upcase!
```

app/validators/waste_carriers_engine/renewal_lookup_validator.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def registration_in_renewal_window?(record, check_service)
+      return true if check_service.in_renewal_window?
+
+      renewable_date = check_service.date_can_renew_from
+      record.errors.add(:temp_lookup_number, :not_yet_renewable, date: renewable_date)
+      false
+    end
+  end
```
cintamani, line 63: I was actually expecting to see this implemented in the validation gem, any particular reason for not doing that? I don't really have a strong opinion about having this here or in the Gem, mostly just curiosity! :D

app/views/waste_carriers_engine/renew_registration_forms/new.html.erb
```diff
@@ -1,5 +1,51 @@
...
 
 <div class="text">
-  <h1 class="heading-large">Renew registration Form - TODO</h1>
+  <%= form_for(@renew_registration_form) do |f| %>
+    <%= render("waste_carriers_engine/shared/errors", object: @renew_registration_form) %>
+
+    <h1 class="heading-large"><%= t(".heading") %></h1>
+
+    <div class="form-group <%= "form-group-error" if @renew_registration_form.errors[:temp_lookup_number].any? %>">
+      <fieldset id="temp_lookup_number">
```
cintamani, line 10: Do we actually need a fieldset here? Me and Andrew have removed a lot of them recently, as they should only be needed when there is more than one field to group together, like in the case of checkboxes or radio buttons :)

spec/validators/waste_carriers_engine/renewal_lookup_validators_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+
+          it "is invalid and sets the correct error message" do
+            expect(subject).to_not be_valid
+            expect(subject.errors[:temp_lookup_number].first).to include("unrenewable_status")
+          end
+        end
+      end
+    end
+  end
+end
```
cintamani, line 135: <3

app/validators/waste_carriers_engine/renewal_lookup_validator.rb
irisfaraway: It's so specific that I didn't think we'd be re-using it in the gem.

app/views/waste_carriers_engine/renew_registration_forms/new.html.erb
irisfaraway: Probably not, I think this just comes from copying and pasting! I'll update it.

PR #768, opened 2020-03-30
Exclude confirmed convictions from :convictions_possible_match scope

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-967

Since `possible_match` is the default value for the convictions workflow state, this meant that old confirmed records were sometimes considered to also be possible matches, clogging up the list of convictions to check.

This PR adds an extra filter to the scope to avoid this happening, and increases our tests for the scopes.


irisfaraway:
Tested this out on pre-prod and seems to work, so now ready for a review.

PR #769, opened 2020-03-31
Fixes to lower tier registration

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-838

This fixes various issue on the lower tier registration created during the new journey.
 - The route was not set when completing the registration, fixed in the state machine
 - The registration finance details were not set. Since the old system generates a fake worldpay payment for an order of 0 balance and no items, we reproduced the same here in the completion job
 - Make sure we don't set an expire date on LT registration

### DIFF
--- a/app/services/waste_carriers_engine/registration_completion_service.rb
+++ b/app/services/waste_carriers_engine/registration_completion_service.rb
@@ -8,11 +8,13 @@ module WasteCarriersEngine
       @transient_registration = transient_registration
 
       transient_registration.with_lock do
+        prepare_finance_details_for_lower_tier
+
         copy_names_to_contact_address
         copy_data_from_transient_registration
 
         set_reg_identifier
-        set_expiry_date
+        set_expiry_date if registration.upper_tier?
 
         update_meta_data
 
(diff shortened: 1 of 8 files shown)


app/services/waste_carriers_engine/registration_completion_service.rb
```diff
@@ -8,11 +8,13 @@ def run(transient_registration)
...
 
       transient_registration.with_lock do
+        prepare_finance_details_for_lower_tier
+
         copy_names_to_contact_address
         copy_data_from_transient_registration
 
         set_reg_identifier
-        set_expiry_date
+        set_expiry_date if registration.upper_tier?
```
irisfaraway: 🤔 Why are we setting the expiry date here, and again in the CanChangeRegistrationService when we activate?

cintamani: yes :) We have been doing that before and that's the same thing that happens on the old service. I spoke to Alan about it few weeks ago and we decided to keep this behaviour. Basically, we always have an expiry date, but we update it when the registration is activated. Same as old service :)

irisfaraway: Ahh, that makes sense! Thanks

PR #786, opened 2020-04-06
Send email when user registers and has pending payment

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-910

If a user submits a new registration and chooses to pay by bank transfer, we should email them with the bank details and confirm that their registration won't be activated until payment is received.

---

See Jira for screenshots!


irisfaraway:
SonarCloud is raging on this, but I'm not sure how valid the comments are for email HTML! I based this all off a template Martin wrote so I trust him to have assessed whether or not it was valid.

PR #788, opened 2020-04-07
Add id to reg identifier spans on completed pages

cintamani:


app/views/waste_carriers_engine/registration_completed_forms/new.html.erb
```diff
@@ -4,7 +4,7 @@
 
     <p class="font-large"><%= t(".highlight_text") %><br>
 
-    <span class="strong"><%= @registration.reg_identifier %></span></p>
+    <span class="strong" id="reg-identifier"><%= @registration.reg_identifier %></span></p>
```
andrewhick, line 7: Most IDs in page objects have underscore rather than hyphen. Could we use `reg_identifier` for consistency?

app/views/waste_carriers_engine/renewal_complete_forms/new.html.erb
```diff
@@ -11,7 +11,11 @@
 
       <div class="govuk-box-highlight">
         <h1 class="heading-xlarge"><%= t(".heading") %></h1>
-        <p class="font-large"><%= t(".highlight_text") %><br><span class="strong"><%= @renewal_complete_form.reg_identifier %></span></p>
+        <p class="font-large">
+          <%= t(".highlight_text") %>
+          <br>
+          <span class="strong" id="reg-identifier"><%= @renewal_complete_form.reg_identifier %></span>
```
andrewhick, line 17: This looks like a mismatch between `-` and `_` here

cintamani: This is just adding new lines to make the code more readable. The only change is the actual addition of the id attribute

PR #797, opened 2020-04-16
Send email when new reg application is pending conviction check

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-842

If a new registration has been paid for but cannot be completed immediately because of convictions checks, we should still send the user a confirmation email to let them know the application was received.

This email is not sent if there is a pending payment at the end of the new registration journey. In that case we send the pending payment email only.

image

### DIFF
--- a/app/services/waste_carriers_engine/registration_completion_service.rb
+++ b/app/services/waste_carriers_engine/registration_completion_service.rb
@@ -73,13 +73,14 @@ module WasteCarriersEngine
       transient_registration.delete
     end
 
+    # Note that we will only send emails here if the registration has pending convictions or pending payments.
+    # In the case when the registration can be completed, the registration activation email is sent from
+    # the RegistrationActivationService.
     def send_confirmation_email
       if registration.unpaid_balance?
         send_pending_payment_email
-        # TODO: Add email for pending convictions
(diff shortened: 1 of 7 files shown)


app/services/waste_carriers_engine/registration_completion_service.rb
```diff
@@ -73,13 +73,14 @@ def delete_transient_registration
...
+    # In the case when the registration can be completed, the registration activation email is sent from
+    # the RegistrationActivationService.
     def send_confirmation_email
       if registration.unpaid_balance?
         send_pending_payment_email
-        # TODO: Add email for pending convictions
-        # Note that we will only send emails here if the registration has pending convictions or pending payments.
-        # In the case when the registration can be completed, the registration activation email is sent from
-        # the RegistrationActivationService.
+      elsif registration.conviction_check_required?
```
cintamani, line 82: I would keep the comment, just not the first `TODO` line.

irisfaraway: It's still there! Just moved it up a few lines.

irisfaraway:
All the Sonarcloud issues are to do with email HTML.

PR #816, opened 2020-04-27
Add new address lookup service and refactor

cintamani:
This adds in the new address lookup service, as per wex, and replace it to the address finder service.
Refactoring was necessary on the bits using it as the old service was not based on our standard BaseService.

Screenshot 2020-04-27 at 13 37 37


spec/support/address_finder_service_stub_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module AddressFinderServiceStubHelper
   # rubocop:disable Metrics/MethodLength
-  def stub_address_finder_service(options)
+  def stub_address_finder_service(options={})
```
cintamani, line 5:
```suggestion
  def stub_address_finder_service(options = {})
```

cintamani:
Holding it on waiting until Alan is around tomorrow - I am not sure how to activate mocking once this is in 😬

PR #820, opened 2020-04-30
Add 'are you sure?' screen when paying by bank transfer

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1004

Users are confused by the current bank transfer flow. An 'are you sure?' page has been added to help users pay by bank transfer.

This replaces the existing bank transfer information page. All details for how to pay by bank transfer will be moved to the page at the end of the journey as part of RUBY-1005.

This only applies to the new registration and renewals journey. Edit and ordering copy cards are not affected.

image


app/forms/waste_carriers_engine/copy_cards_bank_transfer_form.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 module WasteCarriersEngine
-  class CopyCardsBankTransferForm < BankTransferForm
+  class CopyCardsBankTransferForm < BaseForm
```
cintamani, line 4: Why are we changing this? I thought the scope of the ticket wasn't including copy cards flow?

cintamani: Ah! Never mind. I got it :)

PR #822, opened 2020-05-01
Split up 'renewal received' page

irisfaraway:
Currently, if a user renews but the application has a pending payment or convictions check, they will be sent to the renewal received page.

This is a single page, but with a lot of conditional logic about what to display depending on the reason for them seeing it.

In new registrations, we just have two separate pages and everything is a lot clearer.

In preparation for updating the bank transfer info for the 'pending payment' version of this page, this PR splits the page into two separate controllers, routes, views, etc.


irisfaraway:
Think we've agreed not to modify the empty method issues SonarCloud is complaining about.

PR #828, opened 2020-05-05
Add renew via magic link route and function

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-986

Add a renew_token field to the registration, along with a function that will generate a renew link.
Add routes to the magic link renewal process.

### DIFF
--- a/app/controllers/waste_carriers_engine/renews_controller.rb
+++ b/app/controllers/waste_carriers_engine/renews_controller.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class RenewsController < ApplicationController
+    before_action :validate_renew_token
+
+    def new
+      # TODO - Should create a renewing registation
+      # @transient_registration = RenewingRegistration.create(reg_identifier: registration.reg_identifier)
+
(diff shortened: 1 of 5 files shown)


app/controllers/waste_carriers_engine/renews_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class RenewsController < ApplicationController
+    before_action :validate_renew_token
+
+    def new
+      # TODO - Should create a renewing registation
```
cintamani, line 8:
```suggestion
      # TODO: Should create a renewing registation
```

app/controllers/waste_carriers_engine/renews_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      # @transient_registration = RenewingRegistration.create(reg_identifier: registration.reg_identifier)
+
+      # TODO - Should redirect to start renewing registration journey
```
cintamani, line 11:
```suggestion
      # TODO: Should redirect to start renewing registration journey
```

PR #838, opened 2020-05-14
Fix bug on past registration creation at renewal

cintamani:
While working on other tasks, I noticed that the current implementation of PastRegistration#version_already_backed_up?
 is such that if an edit as ever been performed on a registration, at the time of renewal we would miss to create a ne
w past registration for that renewal.

This fixes that situation and add relevant test coverage.

### DIFF
--- a/app/models/waste_carriers_engine/past_registration.rb
+++ b/app/models/waste_carriers_engine/past_registration.rb
@@ -11,9 +11,9 @@ module WasteCarriersEngine
 
     embedded_in :registration, class_name: "WasteCarriersEngine::Registration"
 
-    def self.build_past_registration(registration, cause)
+    def self.build_past_registration(registration, cause = nil)
       past_registration = PastRegistration.new
-      past_registration.cause = cause
+      past_registration.cause = cause if cause.present?
 
       return if past_registration.version_already_backed_up?(registration)
 
@@ -31,7 +31,7 @@ module WasteCarriersEngine
       return false if new_version_due_to_edit?
 
       # Collect all expires_on dates from past registrations
-      matching_expires_on = registration.past_registrations.where(cause: :renewal).map(&:expires_on)
+      matching_expires_on = registration.past_registrations.not_in(cause: ["edit"]).map(&:expires_on)
       # Check if the current expires_on is included - this indicates that this version of
       # the registration has already been backed up.
       return true if matching_expires_on.include?(registration.expires_on)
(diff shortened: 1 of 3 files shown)


spec/models/waste_carriers_engine/past_registration_spec.rb
```diff
@@ -37,18 +37,30 @@ module WasteCarriersEngine
...
-            past_registration
-            expect(registration.reload.past_registrations.count).to eq(past_registration_count)
+          context "if the past registration is not a renewal" do
+            it "does create a new past_registration" do
+              past_registration_count = registration.past_registrations.count
+              past_registration
+              expect(registration.reload.past_registrations.count).to eq(past_registration_count + 1)
+            end
           end
         end
```
irisfaraway, line 66: Do we need an extra test for if the new version is an edit and the past version is a renewal?

app/models/waste_carriers_engine/past_registration.rb
```diff
@@ -31,7 +31,7 @@ def version_already_backed_up?(registration)
       return false if new_version_due_to_edit?
 
       # Collect all expires_on dates from past registrations
-      matching_expires_on = registration.past_registrations.map(&:expires_on)
+      matching_expires_on = registration.past_registrations.where(cause: :renewal).map(&:expires_on)
```
irisfaraway, line 34: What about any existing PastRegistrations which don't have this cause saved? I wonder if we will need to update them, although I think we can at least assume they are renewals.

spec/models/waste_carriers_engine/past_registration_spec.rb
cintamani: I think we do. Is the missing scenario for which this bug slipped into our master I think.

app/models/waste_carriers_engine/past_registration.rb
cintamani: There won't be any new one I have made that param required. 
For past registrations (current, already in the data) with no cause, they won't affect the logic of this method. This is a guard clause that is almost never hit unless we run the RenewlCompletionService multiple times, and that only happen if we have a glitch/issue. Hence, I won't worry about that at all.

Cruikshanks: I think I'm with @irisfaraway on this one. I'm not seeing how on day one of shipping this change that this wouldn't miss a `past_registration` with no cause and a matching expiry date.

cintamani: `miss a past_registration with no cause and a matching expiry date` 
I am saying It would, but is not important. Better to save an extra one (very very few chances of that, as we run this code only once) that to miss one. Happy to have a call whenever :)

PR #839, opened 2020-05-14
Add registration already renewed screen for magic link error

cintamani:
From:https://eaflood.atlassian.net/browse/RUBY-1001

This adds an ad-hoc screen when the renewal link is hit after a registration has already been renewed

Screenshot 2020-05-14 at 13 56 15

### DIFF
--- a/app/models/waste_carriers_engine/registration.rb
+++ b/app/models/waste_carriers_engine/registration.rb
@@ -53,7 +53,10 @@ module WasteCarriersEngine
     end
 
     def already_renewed?
-      check_service.date_can_renew_from > Time.now.in_time_zone("London").to_date
+      period_after_last_window = Rails.configuration.expires_after.years + Rails.configuration.grace_window.months
+      registration_window = expires_on - period_after_last_window + 1.day
+
+      past_registrations.where(cause: nil, :expires_on.gte => registration_window).any?
     end
 
     def expire!
--- a/app/models/waste_carriers_engine/registration.rb
+++ b/app/models/waste_carriers_engine/registration.rb
@@ -52,6 +52,10 @@ module WasteCarriersEngine
       save!
     end
 
+    def already_renewed?
+      check_service.date_can_renew_from > Time.now.in_time_zone("London").to_date
+    end
+
     def expire!
       metaData.status = "EXPIRED"
 
@@ -73,11 +77,14 @@ module WasteCarriersEngine
     end
 
     def renewable_date?
-      check_service = ExpiryCheckService.new(self)
       return true if check_service.in_expiry_grace_window?
       return false if check_service.expired?
 
       check_service.in_renewal_window?
+    end
+
+    def check_service
+      @_check_service ||= ExpiryCheckService.new(self)
     end
   end
 end
--- a/app/services/waste_carriers_engine/expiry_check_service.rb
+++ b/app/services/waste_carriers_engine/expiry_check_service.rb
@@ -4,7 +4,6 @@ module WasteCarriersEngine
   # Contains methods related to dealing with dates in the service, for example
   # whether a date would be considered as expired.
   class ExpiryCheckService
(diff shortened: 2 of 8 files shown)


app/services/waste_carriers_engine/expiry_check_service.rb
```diff
@@ -4,37 +4,38 @@ module WasteCarriersEngine
   # Contains methods related to dealing with dates in the service, for example
   # whether a date would be considered as expired.
   class ExpiryCheckService
-    attr_reader :expiry_date, :registration_date
+    # attr_reader :expiry_date, :registration_date
```
irisfaraway, line 7:
```suggestion
```

config/locales/forms/renews/en.yml
```diff
@@ -0,0 +1,9 @@
+en:
+  waste_carriers_engine:
+    renews:
+      already_renewed:
+        heading: "That registration has already been renewed"
+        paragraph1: "Our records show that registration %{reg_identifier} has already been renewed."
+        paragraph2_html: "If you need to register other waste exemptions you should <a href=\"%{link_to_new_registration}\">make a new registration.</a>"
```
irisfaraway, line 7: Waste exemptions? Think we may need to double-check this

app/models/waste_carriers_engine/registration.rb
```diff
@@ -52,6 +52,10 @@ def generate_renew_token!
       save!
     end
 
+    def already_renewed?
+      check_service.date_can_renew_from > Time.now.in_time_zone("London").to_date
```
irisfaraway, line 56: Won't this also return `true` if the registration is expiring in 4 months from now, and can't be renewed for another month? I feel like this could confuse people.

config/locales/forms/renews/en.yml
cintamani: Ah! I copy-pasted from the wireframe and didn't notice! Will ask for an updated text!

app/models/waste_carriers_engine/registration.rb
cintamani: No, I don't think so. 🤔 I have explored different approaches to decide if a registration was already renewed or not. Every one of them was `flaky` in some part. I choose this at the end as it was the one that made more sense. Happy to discuss the approach tomorrow

Cruikshanks: I believe @irisfaraway is right. Using my local environment I created a situation where I had a registration with an `expires_on` of `2020-09-14` and no `past_registrations`. This put it at 4 months before it expired, and 1 month before it could be renewed.

When I called `already_renewed?` I got `true`.

Screenshot 2020-05-14 at 23 14 17

To try and be helpful I had a stab at coming up with a solution and came up with this.

```ruby
registration_window = reg.expires_on - Rails.configuration.expires_after.years
reg.past_registrations.where(cause: nil, :expires_on.gte => registration_window ).count
```

So I created another registration, this time with an `expires_on` of `2020-06-14`. With nothing in `past_registrations` the query returned `0`. I then completed a renewal and tried again. This time I got `1`.

I then went in and added another registration to `past_registrations` and set its registered on dates to `2014-06-14` and its `expires_on` to `2017-06-14`. So now I have 2 registrations, one covering 2014 to 2017, and the other 2017 to 2020.

Running `reg.past_registrations.where(cause: nil, :expires_on.gte => registration_window ).count` still returns me 1 as expected. If `past_registrations` only contained the 2014-2017 registration, I would not consider the current registration to have been renewed as the past one is too old. I'd expect the query to return `0` (which it does).

So not necessarily the final solution. But hopefully, something that helps.

cintamani: Well, I like it. And using "nil" to refer to past registrations that are renewal will also remove the confusion on the other PR. Will make the changes

PR #840, opened 2020-05-14
Add screen for late renewals

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-1002

This adds an ad-hoc screen for when the magic link is used past the renewal window

Screenshot 2020-05-14 at 14 51 01

### DIFF
--- a/config/locales/forms/renews/en.yml
+++ b/config/locales/forms/renews/en.yml
@@ -1,0 +1,9 @@
+en:
+  waste_carriers_engine:
+    renews:
+      already_renewed:
+        heading: "That registration has already been renewed"
+        paragraph1: "Our records show that registration %{reg_identifier} has already been renewed."
+        paragraph2_html: "If you need to register other waste exemptions you should <a href=\"%{link_to_new_registration}\">make a new registration.</a>"
+        paragraph3: "If you want to discuss this registration, contact the Environment Agency. You will need your registration number."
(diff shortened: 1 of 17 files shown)


config/locales/forms/renews/en.yml
```diff
@@ -7,3 +7,7 @@ en:
         paragraph2_html: "If you still need to register as a waste carrier, broker or dealer then <a href=\"%{link_to_new_registration}\">make a new registration.</a>"
         paragraph3: "If you want to discuss this registration, contact the Environment Agency. You will need your registration number."
         paragraph4_html: "Environment Agency helpline: 03708 506 506 (Monday to Friday, 8am to 6pm) <a href=\"mailto:nccc-carrierbroker@environment-agency.gov.uk\">nccc-carrierbroker@environment-agency.gov.uk</a>"
+      past_renewal_window:
+        heading: "That renewal has expired"
+        paragraph1: "Your registration expired over 1 month ago, so you cannot renew."
```
Cruikshanks, line 12: Had the discussion with @cintamani and we agree this is wrong. Currently, with the grace window being 90 days a registration would have had to expire more than 3 months ago to be ineligible.

And if the grace window comes back down to 5 days, then a registration could be less than 1 month expired and still see this message.

We've agreed to merge for now (everything else is tip-top!) and will raise the content issue with our UCD team.

PR #859, opened 2020-06-05
Remove extra logging and fix order of events

cintamani:
This removes some unnecessary error raised, fix an issue with the order of events and adds an extra check before deciding if a registration can be activated or not.

### DIFF
--- a/app/services/waste_carriers_engine/registration_activation_service.rb
+++ b/app/services/waste_carriers_engine/registration_activation_service.rb
@@ -1,15 +1,14 @@
 # frozen_string_literal: true
 
 module WasteCarriersEngine
-  class UnpaidBalanceError < StandardError; end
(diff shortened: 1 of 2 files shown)


cintamani:
If `can_be_completed?` is false we return rather than raise an error, and this is fine because any finance operation (and not only finance operation) will run that code without having a concern if the registration can actually be activated. That is the unnecessary logging I am removing. 

If `can_be_completed?` is true, then we are saying "this registration is into a state we can activate". Given that an error. with sending an email is already custom logged, we are not worried about that. Other things that can go wrong are going to raise an error that will end up in airbrake for us to see as we are using `!` methods to activate the registration. Hence, I don't think we need more logging. 

Unless I didn't get what you meant, of course, as always XD

PR #861, opened 2020-06-09
Update email validation to match wireframe

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-923

In order to achieve the wireframe validation, we only validate the confirmed email to be equal to the given email and we don't validate its format. There is no need to do so if the format has to be validated for the original email and the two must match.

This achieve having only one validation message that matches what the wireframe shows to the user.


cintamani:
Wow. I really can't think straight today 😮 
Thanks, Iris!

PR #872, opened 2020-06-29
Promote little FeatureToggle to be a proper Model

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-1118

Given that little FeatureToggle has behaved very well in the past few months, we decided to upgrade him to be a proper Model. XD

This model will be used by the new defra-ruby-features engine. Toggles will be persisted in the database.

Things to notice is, in order to handle the fact that the model won't have persisted data about existing toggles, previously persisted in a yml file for read-only, the code will handle that case by redirecting the query for a toggle status to the content of the yml file.
By doing this we will just need to deploy and then take our time creating the toggles through the back office interface. In the meantime, everything will work just fine.

There will need some cleanup to be done at the subsequent deploy, but this still looks like a cleaner solution than a rake task, so that we don't need to cleanup places like Jenkins.

### DIFF
--- a/app/models/waste_carriers_engine/feature_toggle.rb
+++ b/app/models/waste_carriers_engine/feature_toggle.rb
@@ -5,11 +5,20 @@ require "erb"
 
 module WasteCarriersEngine
   class FeatureToggle
+    include Mongoid::Document
+
+    # Use the FeatureToggles database
+    store_in collection: "feature_toggles"
+
+    field :key, type: String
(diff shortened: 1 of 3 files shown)


Cruikshanks:
I intend for [defra-ruby-features](https://github.com/DEFRA/defra-ruby-features) to be referenced as a gem rather than directly from github going forward. But there are a few housekeeping tasks I need to carry out in the gem repo first.

PR #873, opened 2020-07-02
Move feature toggle config to engine

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1118

We have created a number of gems to support reuse of functionality across our services. If the functionality is to be used across both host apps, or because of other functionality, we'll incorporate the gem into the engine. [defra-ruby-alert](https://github.com/DEFRA/defra-ruby-alert) is an example of this.

Our typical pattern when we do this is to then control any config for that gem from the engine, and if extra info is needed from the host apps to pass it via the engines config.

[defra-ruby-features](https://github.com/DEFRA/defra-ruby-features) has been implemented by referencing and configuring the gem directly in the host apps. So to make things consistent this change updates the implementation to match our existing pattern of controlling everything from the engine.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,12 @@ gem "secure_headers", "~> 5.0"
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
 gem "turbolinks"
 
+# Use the Defra Ruby Features gem to allow users with the correct permissions to
+# manage feature toggle (create / update / delete) from the back-office.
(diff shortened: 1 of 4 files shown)


Cruikshanks:
Argh! Just realised this is **not** one of those things we want in the engine and accessible from both apps. This should just be in the back-office. However, there are changes needed in the engine to support that.

So there was a very good reason why this was the way it was in the first place!

PR #881, opened 2020-07-07
Add receipt email to journey end screens pending convictions

irisfaraway:

Users could also get to these pages after paying via WorldPay, so we should include mention of the receipt email on here as well.

Also includes a text update after discussion with Christina.


Cruikshanks:
Updates to the content have been confirmed. It should now be

> A payment confirmation will be sent to [receipt email].

Wireframe has been updated as well.

PR #890, opened 2020-07-15
Add payment email section to payment summary page

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1137

Having removed our first design for solving the problem of capturing a user's preferred Worldpay payment confirmation email address (see [PR #889](https://github.com/DEFRA/waste-carriers-engine/pull/889)) this implements our second version!

That is to show a hidden section in the payment summary screen which will appear when a user selects 'Pay by credit or debit card'. Along with a header and some content, it will display a field, pre-populated with the contact email address. Whatever is in there when submitted will be saved as the `receipt_email`.

This is based on example used by the [Get a fishing licence service](https://get-fishing-licence.service.gov.uk/). In its **How can we send you your licence details?** if the user selects 'Email or text message' a hidden section with those fields will appear.

<details><summary>Get a fishing licence example</summary>

Screenshot 2020-07-15 at 16 45 18

</details>

This change focuses on adding the section and validating it correctly. For example, we only care about the field being populated correctly if the user is paying by card. If they are not then we shouldn't be forcing them to correct any invalid changes they have made to the field.

A subsequent change will focus on the logic of hiding and showing the section at the appropriate times.

<details><summary>Screenshot of the change</summary>

Screenshot 2020-07-16 at 00 06 38

</details>

### DIFF
--- a/app/forms/waste_carriers_engine/payment_summary_form.rb
+++ b/app/forms/waste_carriers_engine/payment_summary_form.rb
@@ -2,7 +2,6 @@
 
 module WasteCarriersEngine
   class PaymentSummaryForm < ::WasteCarriersEngine::BaseForm
-    # delegate :temp_payment_method, to: :transient_registration
 
     attr_accessor :temp_payment_method,
                   :registration_cards,
@@ -18,7 +17,6 @@ module WasteCarriersEngine
     end
 
     def initialize(transient_registration)
-      Rails.logger.debug("----- PAYMENT_SUMMARY_SUBMIT - initialize")
       super
 
       self.temp_payment_method = transient_registration.temp_payment_method
@@ -51,9 +49,6 @@ module WasteCarriersEngine
       self.card_confirmation_email = params[:card_confirmation_email]
     end
 
-    # TODO: This method is duplicated in
-    # app/models/concerns/waste_carriers_engine/can_use_new_registration_workflow.rb.
-    # Consider refactoring to being on the transient_registration itself
     def paying_by_card?
       temp_payment_method == "card"
     end
--- a/app/forms/waste_carriers_engine/payment_summary_form.rb
+++ b/app/forms/waste_carriers_engine/payment_summary_form.rb
@@ -2,39 +2,56 @@
 
 module WasteCarriersEngine
   class PaymentSummaryForm < ::WasteCarriersEngine::BaseForm
-    delegate :temp_payment_method, to: :transient_registration
-    delegate :receipt_email, to: :transient_registration, allow_nil: true
+    # delegate :temp_payment_method, to: :transient_registration
 
-    attr_accessor :type_change, :registration_cards, :registration_card_charge, :total_charge
+    attr_accessor :temp_payment_method, :registration_cards, :registration_card_charge, :total_charge, :card_confirmation_email
 
     validates :temp_payment_method, inclusion: { in: %w[card bank_transfer] }
-    validates :receipt_email, "defra_ruby/validators/email": true, if: :paying_by_card?
-
-    after_initialize :pre_populate_receipt_email
+    validates :card_confirmation_email, "defra_ruby/validators/email": true, if: :paying_by_card?
 
     def self.can_navigate_flexibly?
       false
     end
 
     def initialize(transient_registration)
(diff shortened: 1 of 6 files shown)


Cruikshanks:
For quick reference, Sonarcloud is complaining about

- the `TODO` comment
- the fact the new section in the view contains a fieldset without a legend

The new section was very much based on our `contact_email` form and view, and along with most other pages use of a legend does not appear to be the norm.

If what Sonarcloud is saying is true we need to review this pattern. But I think for now it's better to be consistent and to review this issue at a later date.

spec/forms/waste_carriers_engine/payment_summary_forms_spec.rb
```diff
@@ -51,6 +80,28 @@ module WasteCarriersEngine
...
+
+          context "and the receipt email has been set" do
+            before do
+              payment_summary_form.transient_registration.receipt_email = receipt_email
+            end
+
+            context "to something invalid" do
+              let(:receipt_email) { "foo@bar" }
+
+              it "is is not valid" do
```
irisfaraway, line 92:
```suggestion
              it "is not valid" do
```

spec/forms/waste_carriers_engine/payment_summary_forms_spec.rb
```diff
@@ -51,6 +80,28 @@ module WasteCarriersEngine
...
+                expect(payment_summary_form).not_to be_valid
+              end
+            end
+
+            context "to nothing" do
+              let(:receipt_email) { "" }
+
+              it "is is not valid" do
```
irisfaraway, line 100:
```suggestion
              it "is not valid" do
```

app/forms/waste_carriers_engine/payment_summary_form.rb
```diff
@@ -3,10 +3,14 @@
 module WasteCarriersEngine
   class PaymentSummaryForm < ::WasteCarriersEngine::BaseForm
     delegate :temp_payment_method, to: :transient_registration
+    delegate :receipt_email, to: :transient_registration, allow_nil: true
 
     attr_accessor :type_change, :registration_cards, :registration_card_charge, :total_charge
 
     validates :temp_payment_method, inclusion: { in: %w[card bank_transfer] }
+    validates :receipt_email, "defra_ruby/validators/email": true, if: :paying_by_card?
```
irisfaraway, line 11: I would argue that since we are pre-populating, we should always validate this. If a user selects bank transfer, they will still be submitting a value and we will still be saving it.

If they haven't selected WorldPay first and seen the form, this will always be their already validated contact email, and so there shouldn't be any issues.

However, if they select WorldPay, enter an invalid email and then switch back to bank transfer and submit, we will still be storing that value, only without any validation.

Alternatively I think we should wipe any receipt emails that are submitted in combination with the bank transfer option so we don't store any dodgy data.

Cruikshanks: I had considered this scenario. I was wary of forcing a validation because I didn't want to get into the edge case of trying to force a hidden section to reveal if there is a validation error in a field that is no longer pertinent to them.

I was edging on
- likelihood of it happening
- could always be fixed when in the Edit screen (once I have implemented it)

And hoping no one would notice 😁 

Wiping it for bank transfer then becomes the alternative. But that also has an edge case scenario.

- User registers and pays by card so receipt email stored
- User renews and pays by bank transfer. We wipe the receipt email

Appreciate there is 3 years between these things happening so we would have time to actually implement receipts properly and this design would go away! But never say never!

So my current plan is to go back to my first commit idea and create an attribute on the form, pre-populated by whatever is in the transient_registration.

We then still only validate it if the payment type is card. But we also only pass it to be saved to the transient_registration if the payment type is card.

That way we can properly ignore it if payment type is bank transfer. But we also only overwrite any existing value if they are paying by card.

What do you think?

irisfaraway: That sounds like a good solution to me! :)

irisfaraway:
@Cruikshanks I think I have some working Javascript! Did it as a separate PR here, designed to be merged into your branch: https://github.com/DEFRA/waste-carriers-engine/pull/892

It will need a minor tweak to the host apps to work as well so will get those PRs going on.

PR #891, opened 2020-07-16
Add payment confirmation email to Edit

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1160

Now we have the new `receipt_email` field which we are capturing during new registrations and renewals we need to allow users to edit the value. Else once set (or not) it can never be changed which will cause issues for users who need to change it for subsequent Worldpay payments, for example, copy card orders.


Cruikshanks:
Sonarcloud is complaining about the lack of `<th>` in the table added. This is seen as an accessibility issue. But as commented previously the W3C guidelines state, if a table is being used for layout (which admittedly it shouldn't be) then you should avoid adding `<th>` elements.

PR #892, opened 2020-07-16
Hide receipt email on payment summary unless WorldPay selected

irisfaraway:

In a truly shocking development, this adds some JavaScript to the service.

The purpose of this is to hide the receipt email field from the payment summary page until the "WorldPay" option is selected.

If the page is loaded and "WorldPay" is already selected (eg if the page has been reloaded due to a validation error) then the email field will still be shown.

If "bank transfer" is later selected instead, the field will be hidden again.

Note that this won't work in IE8 and below. The default for these browsers and for disabled JavaScript is to always display the email field.

recording

We will also need to add a little `<%= yield :page_scripts %>` to the application template on the host apps:

- https://github.com/DEFRA/waste-carriers-front-office/pull/574
- https://github.com/DEFRA/waste-carriers-back-office/pull/950

### DIFF
--- a/app/forms/waste_carriers_engine/payment_summary_form.rb
+++ b/app/forms/waste_carriers_engine/payment_summary_form.rb
@@ -2,15 +2,14 @@
 
 module WasteCarriersEngine
   class PaymentSummaryForm < ::WasteCarriersEngine::BaseForm
-    delegate :temp_payment_method, to: :transient_registration
-    delegate :receipt_email, to: :transient_registration, allow_nil: true
-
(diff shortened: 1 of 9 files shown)


app/forms/waste_carriers_engine/payment_summary_form.rb
```diff
@@ -52,5 +51,18 @@ def assign_params(params)
     def paying_by_card?
       temp_payment_method == "card"
     end
+
```
Cruikshanks, line 54: Thing something snuck in when resolving the conflicts or from previous. You can ditch everything from here downwards.

PR #898, opened 2020-07-23
Generate renewal token when one is requested

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1166

If a registration is eligible for renewal NCCC would like us to display the same renewal link that gets included in the renewal reminder emails and letters in registration details.

Currently though we only generate the token used in the link when first generating the reminder email. A registration is eligible for renewal before this is done. And when we first release there will be hundreds that are eligible which will never go through the reminder process because it is too late.

The simplest solution we have for this is to generate it if required. This way we avoid any timing issues or another scheduled task that would have to try and identify all eligible registrations and update them in one go.

So this change will add some logic to automatically generate and save a token if one does not exist when the token is requested.


Cruikshanks:
@irisfaraway checkout https://github.com/DEFRA/waste-carriers-back-office/pull/962 to see how using this version of the engine effects the back-office 😁

Cruikshanks:
More for future reference as finding examples of this was super-hard

- https://stackoverflow.com/questions/7306472/override-a-mongoid-models-setters-and-getters

A subsequent answer recommends using `super()` instead of `self[:renew_token]`. It works to *get* the value in our context, but not to *set* it. As there seemed to be no issues with the implementation I've stuck a pin in this thought!

PR #902, opened 2020-07-29
Don't copy account email around editing

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1178

Previously during an edit, the value of account_email was copied from the registration to the transient_registration and then copied back again when the edit was completed.

If the registration was transferred between the edit being started and completed, this resulted in a bug where the account_email date was "reset" to its previous value.

There was existing code to try and prevent this. But there was a mistake in the naming convention used caused by a legacy issue of the old app not being consistent with camelCase and snake_case field names.

This PR fixes the existing code.

### DIFF
--- a/app/models/waste_carriers_engine/edit_registration.rb
+++ b/app/models/waste_carriers_engine/edit_registration.rb
@@ -9,9 +9,12 @@ module WasteCarriersEngine
 
     validates :reg_identifier, "waste_carriers_engine/reg_identifier": true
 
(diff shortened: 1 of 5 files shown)


Cruikshanks:
> Good spot! Is this an issue around any other transient regs?

I'll be honest, I didn't do an exhaustive check. But a quick review and everything else looks okay. The names look as I recall them to be in the database.

PR #911, opened 2020-09-15
Update privacy policy and declaration for data protection

andrewhick:
This PR updates changes to the privacy policy and declaration, following discussion with our service owner and the Data Protection team. Both are happy with the updates.

It's my first time making a change to the structure of the page, so please especially double check `app/views/waste_carriers_engine/declaration_forms/new.html.erb` and the added text.


andrewhick:
Well spotted Iris, updated :)

irisfaraway (approved):
Approved! 5th seems unlikely for a release now but we can update it once an RFC is in.

PR #974, opened 2021-04-15
Send lower tier confirmation emails with Notify

tobyprivett:
To do
- [x] address formatting
- [x]  attachments
- [x] spec: Notify::RegistrationCompletedService
- [x] fix failing specs
- [x] smoke test with a real email
- [x] remove unused mailer code

Notes:
- This will send Upper and Lower registration certificates via Notify until card #1290 is completed.
- There is no check for the presence of a PDF as we don't actually generate a file (`render_to_string` is used instead). Can add a check if needed, but I'm not sure that it could happen (other than in a fatal error that should be handled by the `RegistrationActivationService`)


app/helpers/waste_carriers_engine/application_helper.rb
```diff
@@ -60,11 +60,6 @@ def displayable_address(address)
        address.country].reject(&:blank?)
     end
 
-    # WickedPdf does come with helpers that you can use
```
irisfaraway, line 63: I'd be happy to drop this if we were only dealing with certificates for emails. However we also allow certificates to be viewed as PDFs in the back office and front office – see https://github.com/DEFRA/waste-carriers-back-office/blob/main/app/controllers/certificates_controller.rb for an example – so might be worth confirming how this change (and adding a `@presenter` variable in the template) will affect it there.

app/services/waste_carriers_engine/notify/registration_activated_service.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  module Notify
+    class RegistrationActivatedService < BaseSendEmailService
```
irisfaraway, line 5: Should we clarify that this is an email-sending service in the name, eg RegistrationActivatedEmailService? Since we're still working out if a confirmation letter is needed as well as the email, there is the chance of us needing a RegistrationActivatedLetterService or something similar in the future.

irisfaraway (changes requested):
This is looking good! Love the addition of a Notify namespace – think that will make things much tidier.

I have some suggested changes, but hopefully nothing too major. I think there's also an I18n locale file for the registration activated email which we can probably delete along with the templates. Happy to discuss all this here or on Slack if you have any questions :)

app/services/waste_carriers_engine/notify/registration_activated_service.rb
```diff
@@ -0,0 +1,44 @@
...
+            last_name: @registration.last_name,
+            phone_number: @registration.phone_number,
+            registered_address: presenter.registered_address_fields.join(", "),
+            date_registered: @registration.metaData.date_registered,
+            link_to_file: Notifications.prepare_upload(pdf)
+          }
+        }
+      end
+
+      def presenter
```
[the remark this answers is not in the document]
tobyprivett: Good idea. Have changed this.

spec/services/waste_carriers_engine/registration_activation_service_spec.rb
```diff
@@ -47,17 +50,6 @@ module WasteCarriersEngine
           expect { service }.to_not change { registration.active? }
         end
       end
-
-      context "when the mailer fails" do
```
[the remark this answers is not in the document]
tobyprivett: Adding a test to check that AirBrake is notified when the service raises an error.

app/services/waste_carriers_engine/notify/registration_activated_service.rb
tobyprivett: Good idea. My initial thought was that this would be assumed by the Notify scope. But Notify isn't just for emails!

app/helpers/waste_carriers_engine/application_helper.rb
tobyprivett: Removing the wicked_pdf helper:
- I spent some time trying to mix in the ApplicationHelper into the service so `wicked_pdf_image_tag` was accessible from the template. Then I realised that the service doesn't actually use wicked_pdf so it felt wrong to load all the helpers for this one method was just formatting an image tag. 
- Have checked with the front office app and the PDF certificate is generated as expected, with the logo.

`@presenter` variable:
- The template already used the instance variable. Have updated the comment on the certificate template

irisfaraway: Aha, that makes sense! Thanks for clarifying :)

irisfaraway (approved):
LGTM! Go ahead and merge whenever you're ready.

PR #975, opened 2021-04-19
Send upper tier confirmation emails with Notify

tobyprivett:
https://eaflood.atlassian.net/jira/software/projects/RUBY/boards/374

### DIFF
--- a/app/services/waste_carriers_engine/notify/registration_activated_email_service.rb
+++ b/app/services/waste_carriers_engine/notify/registration_activated_email_service.rb
@@ -8,7 +8,7 @@ module WasteCarriersEngine
       def notify_options
         {
           email_address: @registration.contact_email,
-          template_id: "889fa2f2-f70c-4b5a-bbc8-d94a8abd3990",
+          template_id: template_id,
           personalisation: {
             reg_identifier: @registration.reg_identifier,
             first_name: @registration.first_name,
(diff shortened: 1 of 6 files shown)


app/services/waste_carriers_engine/notify/registration_activated_email_service.rb
```diff
@@ -8,10 +8,9 @@ class RegistrationActivatedEmailService < BaseSendEmailService
       def notify_options
         {
           email_address: @registration.contact_email,
-          template_id: "889fa2f2-f70c-4b5a-bbc8-d94a8abd3990",
+          template_id: template_id,
           personalisation: {
             reg_identifier: @registration.reg_identifier,
-            registration_type: @registration.registration_type,
```
irisfaraway, line 14: I think this is an issue with the Notify template rather than the code, as it's apparently missing the template variable, but it looks like where it says "You are now registered as an upper tier waste carrier and dealer", this should be where we display the registration type. This could be "carrier and dealer", "broker and dealer" or "carrier, broker and dealer".

Lower tier regs don't have to select a registration type, so it wasn't an issue there.

irisfaraway (changes requested):
Looks good :) Nice and straightforward!

I spotted one issue with the template which I think will mean we still want to have the `registration_type` in the personalisation, so that's the only reason I'm not hitting approve yet – let me know what you think.

app/services/waste_carriers_engine/notify/registration_activated_email_service.rb
tobyprivett: We might be better off with a stringified registration type that can be added into the template. I will look into it.

irisfaraway (approved):
LGTM! :shipit:

PR #980, opened 2021-04-22
Send 'registration received pending payment' emails from Notify

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1347


app/services/waste_carriers_engine/notify/base_send_email_service.rb
```diff
@@ -3,6 +3,9 @@
 module WasteCarriersEngine
   module Notify
     class BaseSendEmailService < BaseService
+      include WasteCarriersEngine::ApplicationHelper
```
tobyprivett, line 6: @irisfaraway Mixed into the base class. Do you think that's the best place or would it sit better in the class that uses it?

irisfaraway: If it gets reused by multiple child classes (which I think it does?) then I'm fine with having it in here.

spec/cassettes/notify_registration_pending_payment_sends_an_email.yml
```diff
@@ -0,0 +1,69 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://api.notifications.service.gov.uk/v2/notifications/email
+    body:
+      encoding: UTF-8
+      string: '{"email_address":"foo@example.com","template_id":"b8b68a4c-adc9-4fe6-86cd-3d5a83822c47","personalisation":{"reg_identifier":"CBDU1","first_name":"Jane","last_name":"Doe","sort_code":"","account_number":"","payment_reference":"","payment_due":"","iban":"","swift_bic":"","currency":""}}'
```
irisfaraway, line 8: It looks like this is missing all the finance parameters, both the payment due and the I18n ones – not sure what's going on here as the specs for the RegistrationPendingPaymentEmailService do appear to be checking that the expected parameters are received, and the test is passing.

It's possible this is just an outdated VCR cassette, although I would have expected VCR to regenerate it anyway since a test with all the personalisation data would have generated a different request.

Can we maybe try re-generating this cassette or something and see if that fixes it?

irisfaraway (commented):
This is looking good!

I have one comment about one of the VCR cassettes – it's missing some of the data I'd expect to see in the request but the tests are still passing.

spec/cassettes/notify_registration_pending_payment_sends_an_email.yml
tobyprivett: I've rebuilt the VCR cassettes and pushed an update.

Strange behaviour, though. Have tested locally with no internet connection and changed the notify parameters and it behaves expected.

irisfaraway (approved):
:shipit:

PR #983, opened 2021-04-29
Send registration card order confirmation emails from Notify

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1374


app/services/waste_carriers_engine/notify/base_send_email_service.rb
```diff
@@ -6,8 +6,9 @@ class BaseSendEmailService < BaseService
       include WasteCarriersEngine::ApplicationHelper
       include ActionView::Helpers::NumberHelper
 
-      def run(registration:)
+      def run(registration:, order: nil)
```
tobyprivett, line 9: I *think* it is okay to have the order param in the base class. Happy to change it, though.

irisfaraway (approved):

PR #988, opened 2021-05-05
Send 'renewal received pending conviction check' emails from Notify

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1392

- Emails are sent on a change of state.

- The logic to determine which template to use was in the ReminderMailer. This
won't work any more, as we are replacing the mailer with a single-responsibility
Notify class.

- Here we specify the notify service/template in the state machine.
I hope it works.


spec/models/waste_carriers_engine/registration_spec.rb
```diff
@@ -604,7 +604,8 @@ module WasteCarriersEngine
             registration.metaData.activate!
 
             # Use .to_i to ignore milliseconds when comparing time
-            expect(registration.reload.expires_on.to_i).to eq(3.years.from_now.to_i)
+            expect(registration.reload.expires_on.to_i)
```
tobyprivett, line 607: This spec fails on this PR. There is a one second difference between the two dates. Can you spot where I might have introduced this?

irisfaraway (approved):
This looks good – I definitely prefer the state machine `success:` actions all being separate methods.

Not sure where that 1 second difference is coming from! It may just be that the Notify service takes slightly longer than the old mailer. But a second off still seems accurate enough to me.

PR #994, opened 2021-05-10
Send Upper tier AD confirmation letter automatically through Notify

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1286


app/services/waste_carriers_engine/registration_confirmation_service.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class RegistrationConfirmationService < BaseService
```
irisfaraway, line 4: Really like this! 🙌

irisfaraway (approved):
LGTM! Looks like it'll be straightforward to get the lower tier version sorted as well :)

PR #1056, opened 2021-11-24
Address reuse for non-corporate users

tobyprivett:
Address reuse for non-corporate users

- Here we add an interstitial page to the workflow:
"Do you want to use this address
as a contact address?"

- When a user selects 'yes', the contact address is populated and they skip
the contact address forms

- When a user selects 'no', they continue along the contact address journey

https://eaflood.atlassian.net/browse/RUBY-1697


Cruikshanks (approved):
It's been almost 18 months since I've had to deal with WEX and WCR but as far as I can see this all fits in and makes sense from a general ruby/rails point of view.

So, good job. Merge away! 👍

PR #1075, opened 2021-12-31
Implement companies house data reuse for registration

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1732

Here we introduce a new workflow that uses Companies House data to populate company/details.

- Applies to a Limited company / LLP and an UPPER tier registration,
- For a lower tier registration, we do not currently ask a user for their company number, so this new workflow does not apply.


tobyprivett:
Closing this as we are reconsidering the Companies House integration

PR #1086, opened 2022-01-20
Feature/page title match h1

Beckyrose200:
https://eaflood.atlassian.net/jira/software/projects/RUBY/boards/374?selectedIssue=RUBY-1723
Here we have changed the titles tags to match the page H1 tags to make the text more accessible for those using a screen reader.

The only exception is on config/locals/forms/waste_types_forms/en.yml where the header doesn't provide enough context for the user so the page title has been kept in place instead.

The changes come as part of the accessibility audit.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
PaulDoyle-DEFRA, line 1: Add .vscode to .gitignore so that your settings are not committed.

PaulDoyle-DEFRA (approved):

PR #1093, opened 2022-02-07
Feature/order items log

PaulDoyle-DEFRA:
This change adds a reporting table for order items, and populates it when registrations are activated or renewed, and when copy cards are ordered mid-registration.
The new collection will be queried by waste-carriers-back-office to generate order reports.

### DIFF
--- a/app/models/waste_carriers_engine/order_item_log.rb
+++ b/app/models/waste_carriers_engine/order_item_log.rb
@@ -13,6 +13,7 @@ module WasteCarriersEngine
     field :activated_at,    type: DateTime
     field :type,            type: String
     field :quantity,        type: Integer
+    field :exported,        type: Boolean, default: false
 
     def self.create_from_registration(registration)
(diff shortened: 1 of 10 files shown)


app/models/waste_carriers_engine/order_item_log.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
```
tobyprivett, line 1: We have this in the back office PR. Should really be in one place or the other. It's scope to `WasteCarriersEngine` so the engine would make most sense

tobyprivett (commented):
Looks good! Just a comment about the duplicate file

app/models/waste_carriers_engine/order_item_log.rb
PaulDoyle-DEFRA: Agreed, this can stay in the engine and I will remove it from the back-office.

PR #1102, opened 2022-02-16
Feature/log mid reg card orders

PaulDoyle-DEFRA:
This change uses the current date/time as the activation timestamp for order_item_logs created as a result of a mid-registration (ad-hoc) copy cards order. Previously the registration activation date/time was used, but that is not correct for a mid-registration card order.


app/services/waste_carriers_engine/order_copy_cards_completion_service.rb
```diff
@@ -31,7 +31,7 @@ def update_registration
 
       # Log the order items only if payment is complete.
       if !@transient_registration.unpaid_balance?
-        OrderItemLog.create_from_registration(registration)
+        OrderItemLog.create_from_registration(registration, Time.now)
```
tobyprivett, line 34: We had some issues recently with Times/Zones. Could we use `Time.current` instead? https://apidock.com/rails/v5.2.3/DateTime/current/class

tobyprivett (changes requested):
Looking good! Just a comment about time zones. We can discuss if you like.

tobyprivett (approved):

PR #1106, opened 2022-02-24
Improve legal entities data

PaulDoyle-DEFRA:
Epic: https://eaflood.atlassian.net/browse/RUBY-1668

This change:
- Adds a registered_company_name attribute to registrations, which is to be populated with the name retrieved via the companies house API.
- Renames the registrations company_name attribute to business_name, to reflect the agreed terminology.
- Adds a company_name presentation helper method to handle the logic of whether either/both names are present and whether to concatenate them in the form "Registered company name trading as business name".


app/services/waste_carriers_engine/notify/registration_confirmation_letter_service.rb
```diff
@@ -39,7 +39,7 @@ def base_personalisation
         {
           contact_name: contact_name,
           reg_identifier: @registration.reg_identifier,
-          company_name: @registration.company_name,
+          business_name: @registration.business_name,
```
tobyprivett, line 42: Notify template would need updating to support this parameter

PaulDoyle-DEFRA: Following discussion - too many potential downstream impacts, reverting the database attribute name change.

PR #1109, opened 2022-03-01
Copy card improvements - stop non card payment ad-hoc card orders

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1791

Here we disable the 'bank_transfer' payment option for a copy card order
process in the back office.

We don't have a good mechanism for testing the state of page objects
within the engine, but this acceptance test will need fixing:

https://github.com/DEFRA/waste-carriers-acceptance-tests/blob/main/features/bo/dashboard/order_cards.feature#L29

### DIFF
--- a/app/views/waste_carriers_engine/copy_cards_payment_forms/new.html.erb
+++ b/app/views/waste_carriers_engine/copy_cards_payment_forms/new.html.erb
@@ -54,7 +54,10 @@
             </div>
 
             <div class="govuk-radios__item">
-              <%= f.radio_button :temp_payment_method, "bank_transfer", class: "govuk-radios__input" %>
+              <%= f.radio_button :temp_payment_method,
+                  "bank_transfer",
+                  class: "govuk-radios__input",
+                  disabled: WasteCarriersEngine.configuration.host_is_back_office? %>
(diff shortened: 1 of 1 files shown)


app/views/waste_carriers_engine/copy_cards_payment_forms/new.html.erb
```diff
@@ -54,7 +54,10 @@
             </div>
 
             <div class="govuk-radios__item">
-              <%= f.radio_button :temp_payment_method, "bank_transfer", class: "govuk-radios__input" %>
+              <%= f.radio_button :temp_payment_method,
+                  "bank_transfer",
+                  class: "govuk-radios__input",
+                  disabled: WasteCarriersEngine.configuration.host_is_back_office? %>
```
PaulDoyle-DEFRA, line 60: Is the `host_is_back_office?` check necessary, as this view is only for back-office use?

PaulDoyle-DEFRA (approved):

app/views/waste_carriers_engine/copy_cards_payment_forms/new.html.erb
tobyprivett: This is only to be disabled when used in the back office. I'm not sure this page ever gets hit in the front office, but it's in the engine so it could

PR #1111, opened 2022-03-02
Enable `active_registration` for an OrderItemLog

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1787

- Before exporting a copy card item, we need to check that the
registration is active

- Here we mixin the `CanCheckRegistrationStatus` concern
and declare that an `OrderItemLog belongs_to Registration`

- And we update the existing specs to ensure that the `Registration`
is persisted before it is assigned to the `OrderItemLog`

### DIFF
--- a/app/models/waste_carriers_engine/order_item_log.rb
+++ b/app/models/waste_carriers_engine/order_item_log.rb
@@ -16,9 +16,11 @@ module WasteCarriersEngine
     field :quantity,        type: Integer
     field :exported,        type: Boolean, default: false
 
-    # enable the CanCheckRegistrationStatus concern
     belongs_to :registration
-    delegate :metaData, to: :registration
+
+    def active_registration?
+      registration.active?
+    end
 
     def self.create_from_registration(registration, activation_time = nil)
       registration.finance_details.orders.each do |order|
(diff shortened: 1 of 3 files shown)


app/models/waste_carriers_engine/order_item_log.rb
```diff
@@ -6,6 +6,7 @@
 module WasteCarriersEngine
   class OrderItemLog
     include Mongoid::Document
+    include CanCheckRegistrationStatus
 
     field :registration_id, type: BSON::ObjectId
```
PaulDoyle-DEFRA, line 11: I think this can be dropped now as `belongs_to :registration` will make the id available.

PaulDoyle-DEFRA (dismissed):

app/models/waste_carriers_engine/order_item_log.rb
```diff
@@ -6,6 +6,7 @@
 module WasteCarriersEngine
   class OrderItemLog
     include Mongoid::Document
+    include CanCheckRegistrationStatus
```
PaulDoyle-DEFRA, line 9: Can this be dropped now that `active_registration?` has been added?

PaulDoyle-DEFRA (dismissed):

app/models/waste_carriers_engine/order_item_log.rb
tobyprivett: Yes. Removing...

PR #1140, opened 2022-03-22
Registered Company Name saved to the transient registration

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1745

Here we have the work for confirming the registered company name. Once the user has confirmed the name and address that is pulled from the company house API (using their inputted company number) we are then saving their registered company name into the transient registration. This populates in the database as registeredCompanyName. This leaves the field companyName in the database which is optional for the user to enter a trading/business name

To Do:
* Testing

### DIFF
--- a/app/forms/waste_carriers_engine/check_registered_company_name_form.rb
+++ b/app/forms/waste_carriers_engine/check_registered_company_name_form.rb
@@ -3,15 +3,22 @@
 module WasteCarriersEngine
   class CheckRegisteredCompanyNameForm < ::WasteCarriersEngine::BaseForm
     delegate :company_no, to: :transient_registration
+    delegate :registered_company_name, to: :transient_registration
     delegate :temp_use_registered_company_details, to: :transient_registration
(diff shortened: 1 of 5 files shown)


app/forms/waste_carriers_engine/check_registered_company_name_form.rb
```diff
@@ -3,17 +3,24 @@
 module WasteCarriersEngine
   class CheckRegisteredCompanyNameForm < ::WasteCarriersEngine::BaseForm
     delegate :company_no, to: :transient_registration
+    delegate :registered_company_name, to: :transient_registration
     delegate :temp_use_registered_company_details, to: :transient_registration
     validates :temp_use_registered_company_details, "waste_carriers_engine/yes_no": true
 
-    def company_name
+    def registered_company_name
```
tobyprivett, line 10: Do we have any tests around this?

tobyprivett (changes requested):

tobyprivett (approved):

PR #1143, opened 2022-03-23
Do not deduplicate "trading as" for lower-tier entities

PaulDoyle-DEFRA:
This change is to skip the removal of "trading as" detail from business name for lower-tier entities.
To avoid over-complicating the code and unit tests, it introduces a new legal_entity_name method for registrations, and uses this to streamline the logic for the entity_display_name.  The associated unit tests have been restructured to more clearly reflect the expected behaviour.
In addition, the presenter "concern" CanPresentEntityDisplayName is moved to a model concern. This consolidates entity naming logic in the model layer and should simplify presenter logic. 
https://eaflood.atlassian.net/browse/RUBY-1805
https://eaflood.atlassian.net/browse/RUBY-1830


spec/support/shared_examples/can_have_registration_attributes.rb
```diff
@@ -402,4 +402,88 @@
...
+      build(factory,
+            business_type: business_type,
+            tier: tier,
+            registered_company_name: registered_company_name,
+            key_people: key_people)
+    end
+
+    subject { resource.legal_entity_name }
+
+    shared_examples "returns registered name" do
```
tobyprivett, line 422: Can we clarify that this refers to the `registered_company_name` and that business name refers to the `company_name`

spec/support/shared_examples/can_present_entity_display_name.rb
```diff
@@ -1,101 +1,113 @@
...
-      end
+  shared_examples "simply legal_entity_name" do
+    it "returns the legal entity name" do
+      expect(subject).to eq registration.legal_entity_name
     end
+  end
 
-    context "with a business name" do
-      let(:trading_as_name) { Faker::Company.name }
+  shared_examples "with a business name with truncation" do
```
tobyprivett, line 37: Can we clarify what a 'business name'  is in this instance? I think it's a `company_name` ?

tobyprivett (dismissed):
Just a comment about 'business name'. Other than that, should be good to go

tobyprivett (approved):

PR #1145, opened 2022-03-24
intermittent unit test failure on check_registered_company_name_form

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1826

Here we have changed the check_registered_company_name test to include html escape as the test was failing if the companies name included an apostrophe.
This change has been tested using a hard coded company name including an apostrophe among other symbols. 
Once happy the test was passing we then changed the test back to including Faker


tobyprivett (approved):

PaulDoyle-DEFRA:
This unescapes the whole response page. I think it would be better to just escape company_name in the comparison.

PR #1155, opened 2022-04-07
Refactor worldpay card payments

PaulDoyle-DEFRA:
These changes are a preparatory step for migration from Worldpay to GovPay. The aim is to separate general online-payments logic from Worldpay-specific logic.
Changes include:
* Rename classes referencing “worldpay” which are about online payments in general and not worldpay specifically.
* Rename keys (not values) which reference worldpay in locale files.
Changes to the following are specifically avoided:
* Names of worldpay-specific classes.
* Database schema (implied) or contents - i.e. the database still has contents such as `payment_status: "WORLDPAY_MISSED"`.
* Environment settings
* Locale file values (as opposed to keys)
* Anticipatory refactoring based on assumptions about how GovPay integration will be implemented.
https://eaflood.atlassian.net/browse/RUBY-1848


app/models/concerns/waste_carriers_engine/can_have_payment_type.rb
```diff
@@ -42,11 +42,11 @@ def bank_transfer?
         payment_type == BANKTRANSFER
       end
 
-      def worldpay?
+      def online?
```
tobyprivett, line 45: This gets tricky because when the conditional method is called `worldpay?`, it reflects the `payment_type`. Could this be misleading for the next people?

PR #1160, opened 2022-04-12
Reset invalid renewal workflow_state value before resuming a renewal

PaulDoyle-DEFRA:
This change addresses an issue which arose when removing the `tier_check` sub-flow from the WCR renewals journey.

If a renewal was in progress prior to the deployment of that change and was paused with one of the `tier_check` `workflow_state` values, and the renewal was resumed after the deployment, the state would be invalid and the renewal would not be able to proceed. 

This change checks that the `workflow_state` is valid  - specifically, that a `next` transition is valid - and if so, updates the renewing registration's `workflow_state` to a valid value.

https://eaflood.atlassian.net/browse/RUBY-1851

### DIFF
--- a/spec/requests/waste_carriers_engine/renewal_start_forms_spec.rb
+++ b/spec/requests/waste_carriers_engine/renewal_start_forms_spec.rb
@@ -77,10 +77,12 @@ module WasteCarriersEngine
                   end
                 end
 
-                it "redirects to the form for the current state" do
-                  get new_renewal_start_form_path(transient_registration[:token])
+                context "when the token is a secure token" do
+                  it "redirects to the form for the current state" do
+                    get new_renewal_start_form_path(transient_registration[:token])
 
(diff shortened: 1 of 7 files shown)


spec/requests/waste_carriers_engine/renewal_start_forms_spec.rb
```diff
@@ -83,6 +83,29 @@ module WasteCarriersEngine
...
+
+                context "when the token is a reg_identifier" do
+                  it "redirects to the location form" do
+                    get new_renewal_start_form_path(transient_registration.registration.reg_identifier)
+
+                    expect(response).to redirect_to(new_location_form_path(transient_registration[:token]))
+                  end
+                end
+
+                it "redirects to the location form" do
```
tobyprivett, line 103: Could this be placed in a separate context from the above `it "redirects to the location form"` to clarify between these two tests

tobyprivett (changes requested):

tobyprivett (approved):

PR #1189, opened 2022-05-26
Remove companies house number validation from check answers form

PaulDoyle-DEFRA:
The renewals check_your_answers form validates the format of companies house number, a relic from before the Companies House API was used to validate the company details. This validation fails for a (valid) 7-digit companies house number.
This change removes that redundant validation. It also removes the redundant check on whether the registration's companies house number has been changed, as the user has no opportunity to change it in the current journey.
https://eaflood.atlassian.net/browse/RUBY-1939


d-a-v-e (approved):
looks good to me - always refreshing to clear out redundant code!

PR #1191, opened 2022-06-01
Chore/back nav new

PaulDoyle-DEFRA:
The current solution for controlling back-navigation in the front- and back-office apps uses the same mechanism as forward-navigation: A state-machine-based workflow definition consisting of a set of permitted transitions, many with conditions based on previous navigation choices by the user. This is complex to maintain and unit test. Recent changes to form flows have exacerbated this.

This change aims to simplify the back-navigation mechanism without modifying the user experience. It does this by maintaining an ordered list of previous forms visited, adding to the list as the user progresses through the forms. When the user navigates back, they are brought to the last form in the list, which is then removed from the list.

In the event that there is an invalid entry in the list, that entry will be skipped, so the user could in theory be brought back more than one step in the flow. If no valid entry is found in the list, they will be brought back to the start form. This is assumed to be acceptable as it should only happen in rare cases, e.g. changes to form navigation impacting a transient registration that was created before the deployment of the change and continued afterwards.

Note that the bulk of the changes are for the removal of the existing hard-wired back-navigation unit tests. The key files for the new mechanism are:
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/config/routes.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/app/controllers/waste_carriers_engine/forms_controller.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/app/controllers/waste_carriers_engine/worldpay_forms_controller.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/app/models/concerns/waste_carriers_engine/can_use_new_registration_workflow.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/app/models/concerns/waste_carriers_engine/can_use_renewing_registration_workflow.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/app/models/concerns/waste_carriers_engine/can_use_edit_registration_workflow.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/app/models/concerns/waste_carriers_engine/can_use_order_copy_cards_workflow.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/app/models/waste_carriers_engine/transient_registration.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/spec/models/waste_carriers_engine/transient_registration_spec.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/spec/requests/waste_carriers_engine/transient_registrations_spec.rb
- https://github.com/DEFRA/waste-carriers-engine/blob/chore/back_nav_new/app/views/waste_carriers_engine/shared/_back.html.erb
- All the form views (changes to the path helper in the back-navigation link)
- The registration, renewal and edit completion services (exclude the new workflow_history attribute when copying details to the registration)


kennyevil (approved):

Beckyrose200:
Can't see any issues with the tests at all, all looks good :)

PR #1217, opened 2022-07-07
[1995] Ruby 1995 wcr bo registration fails to load when payment method is nil

d-a-v-e:
https://eaflood.atlassian.net/browse/RUBY-1995

This adds the missing attributes (`:payment_method`, `:govpay_status`) when creating and order


.rubocop_todo.yml
```diff
@@ -0,0 +1,12 @@
+# This configuration was generated by
```
PaulDoyle-DEFRA, line 1: Would it be better to stick with the standard DEFRA ruby style and to use `rubocop:disable` for the specific cases where the complexity is above the threshold? This change raises the threshold for the entire codebase.

d-a-v-e: done 👍

PR #1224, opened 2022-07-13
Contact email optional

PaulDoyle-DEFRA:
This change allows contact email to be optional in the back-office only. It adjusts validation on the contact-email and check-your-answers forms accordingly.
This is the first step in removing the dependency on a specific NCCC email address to determine whether to send letters or emails to the carrier/dealer.
https://eaflood.atlassian.net/browse/RUBY-1720


spec/validators/waste_carriers_engine/contact_email_validator_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      let(:no_contact_email) { nil }
+
+      it_behaves_like "contact email address is required"
+    end
+
+    context "when running in the back office" do
+      before { allow(WasteCarriersEngine.configuration).to receive(:host_is_back_office?).and_return(true) }
+
+      context "with no_contact_email set to zero" do
+        let(:no_contact_email) { "0" }
```
Beckyrose200, line 75: I don't understand the use of the numbers 0, and 1. Can you point me in the right direction

kennyevil (approved):

spec/validators/waste_carriers_engine/contact_email_validator_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        let(:no_contact_email) { nil }
+        it_behaves_like "contact email address is required"
+      end
+
+      context "with no_contact_email set to one" do
+        let(:no_contact_email) { "1" }
+
+        context "with an email address" do
+          let(:contact_email) { Faker::Internet.email }
+          it_behaves_like "is not valid"
```
Beckyrose200, line 89: Why would it be invalid with a email address present? Is that if they've ticked the box to say no email?

PaulDoyle-DEFRA: Yes, there is a specific requirement to *not* allow the contact email address to be populated if the "no email" checkbox has been ticked.

Beckyrose200: Makes sense, all went straight over my head to begin with

PR #1232, opened 2022-07-21
correct missing translation

d-a-v-e:
bugs raised in:
https://eaflood.atlassian.net/browse/RUBY-1901
https://eaflood.atlassian.net/browse/RUBY-1979

### DIFF
--- a/app/controllers/waste_carriers_engine/govpay_forms_controller.rb
+++ b/app/controllers/waste_carriers_engine/govpay_forms_controller.rb
@@ -55,6 +55,7 @@ module WasteCarriersEngine
         redirect_to_correct_form
       else
         log_and_send_govpay_response(false, action)
+        action = GovpayPaymentDetailsService.payment_status(action)
         flash[:error] = I18n.t(".waste_carriers_engine.govpay_forms.#{action}.invalid_response")
         go_back
       end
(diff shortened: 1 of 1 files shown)


app/controllers/waste_carriers_engine/govpay_forms_controller.rb
```diff
@@ -55,6 +55,7 @@ def respond_to_acceptable_payment(action)
         redirect_to_correct_form
       else
         log_and_send_govpay_response(false, action)
+        action = GovpayPaymentDetailsService.payment_status(action)
```
PaulDoyle-DEFRA, line 58: Does this line do anything?

d-a-v-e: there's a class method in `GovpayPaymentDetailsService` that maps some actions which match up with the locales currently defined.

```ruby
def self.payment_status(status)
  {
    "created" => :pending,
    "started" => :pending,
    "submitted" => :pending,
    "cancelled" => :cancel,
    "failed" => :failure,
    nil => :error
  }.freeze[status] || status.to_sym
end
```

PaulDoyle-DEFRA (approved):

PR #1239, opened 2022-07-27
Add debugging for intermittent production nil user error

PaulDoyle-DEFRA:
This adds debugging for an intermittent production error, which occurs during the checking of permissions for a renewing registration. It has not been possible to replicate the error in a local development environment, hence this temporary additional debugging information.
https://eaflood.atlassian.net/browse/RUBY-1921


timstone:
@PaulDoyle-DEFRA if you want to deploy this code to production, do you want to turn it on and off with a feature toggle?

timstone:
Also it might be helpful to know tier and if the registration has an account

d-a-v-e (dismissed):

kennyevil (approved):

timstone:
> @PaulDoyle-DEFRA if you want to deploy this code to production, do you want to turn it on and off with a feature toggle?

No, I don't think that's necessary. It was just an idea, if we are going to capture errrors then we should have the best available information to troubleshoot.

PaulDoyle-DEFRA:
Replaced by https://github.com/DEFRA/waste-carriers-engine/pull/1249

PR #1249, opened 2022-08-09
Production debug logging

PaulDoyle-DEFRA:
This change adds a general-purpose concern for writing transient registration details to the Rails log and Errbit. It is intended to help debug some intermittent errors which occur in production but with insufficient details in the logs to resolve them.
Note that this PR replaces an earlier version.
https://eaflood.atlassian.net/browse/RUBY-1921


Beckyrose200 (approved):
Looks good

PR #1251, opened 2022-08-10
[2012] Remove logging for govpay

d-a-v-e:
https://eaflood.atlassian.net/browse/RUBY-2012


PaulDoyle-DEFRA (changes requested):
This debug-level logging could possibly stay - it might be useful for capturing Govpay responses in non-production environments.  The offending log messages as per https://eaflood.atlassian.net/browse/RUBY-2012?focusedCommentId=425156 are the "Valid Govpay response" messages being sent from `GovpayFormsController`, in the `log_and_send_govpay_response` method.

PR #1262, opened 2022-09-02
Adding in telephone to the search terms

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-2074

Here we have the changes in the engine to search for a telephone number. The search functionality is in the BO however the service to search is in the engine.

### DIFF
--- a/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
+++ b/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
@@ -92,6 +92,7 @@ module WasteCarriersEngine
                { company_name: /#{escaped_term}/i },
                { last_name: /#{escaped_term}/i },
                { registered_company_name: /#{escaped_term}/i },
+               { phone_number: /#{escaped_term}/i },
                "addresses.postcode": /#{escaped_term}/i)
       }
 
--- a/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
+++ b/app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
@@ -88,16 +88,24 @@ module WasteCarriersEngine
       scope :search_term, lambda { |term|
         if term.present?
           escaped_term = Regexp.escape(term)
+          # Remove any non-digits excluding "+"
           telephone_number = term.gsub(/[^+\d]/, "")
+
+          # Removing the 0 or +44 at the beginning of the number as this is already included in the regex
+          # For numbers not starting in either of these the regex will still work as the 0 and +44 is optional
           if telephone_number.start_with?("+44")
             telephone_number.gsub!("+44", "")
           elsif telephone_number.start_with?("0")
             telephone_number.slice!(0)
           end
 
+          # Avoid trivial matches with search terms intended for other attributes
+          # e.g. search "WEX01" reduces to "01" matching all phone numbers with a "01"
+          # Also, searching with an empty value would match all model instances
           regex = if telephone_number.length < 10
                     escaped_term
                   else
+                    # Regex can search for a number with spaces and dashes anywhere and for UK numbers either starting in 0 or +44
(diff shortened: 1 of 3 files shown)


app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -86,12 +86,27 @@ module CanHaveRegistrationAttributes
       field :total_fee,                                        type: String
 
       scope :search_term, lambda { |term|
-        escaped_term = Regexp.escape(term) if term.present?
+        if term.present?
```
irisfaraway, line 89: Swapping this for `return unless term.present?` could reduce the indentation here

spec/support/shared_examples/search_scope.rb
```diff
@@ -6,48 +6,47 @@
            :has_required_data)
   end
 
+  shared_examples "a matching and a non matching record" do
```
irisfaraway, line 9: Nice :)

spec/support/shared_examples/searching_phone_number.rb
```diff
@@ -0,0 +1,110 @@
...
+  let(:normal_number) { "01234567890" }
+  let(:number_with_spaces) { "012 3456 7890" }
+  let(:number_with_dashes) { "012-3456-7890" }
+  let(:number_starting_with_44) { "+441234567890" }
+  let(:interntational_number) { "+78121234567" }
+  let(:number_with_text) { "Landline 01234567890" }
+
+  context "when the number in the database has not got any spaces or dashes and doesn't start in +44" do
+    context "and the search term has not got any spaces or dashes and doesn't start in +44" do
+      let(:term) { normal_number.to_s }
```
irisfaraway, line 35: Do we need all the `.to_s`s here? Looks like they're all strings already

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -86,12 +86,27 @@ module CanHaveRegistrationAttributes
...
+          if telephone_number.start_with?("+44")
+            telephone_number.gsub!("+44", "")
+          elsif telephone_number.start_with?("0")
+            telephone_number.slice!(0)
+          end
+
+          regex = if telephone_number.length < 10
+                    escaped_term
+                  else
+                    "(\\+44|0|\\+)?[\\s-]*" + telephone_number.scan(/\d/).map { |c| "#{c}[\\s-]*" }.join
```
irisfaraway, line 101: I think this regex would benefit from a comment explaining what is going on

irisfaraway (commented):
This looks good! I like the test refactor. Have some minor suggestions, happy to discuss 👍

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
irisfaraway, line 89: Ignore what I said previously about the `return` clause! I do think this block would benefit from being moved to a different method – `process_phone_number_term` or something like that.

app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -92,9 +92,33 @@ module CanHaveRegistrationAttributes
                { company_name: /#{escaped_term}/i },
                { last_name: /#{escaped_term}/i },
                { registered_company_name: /#{escaped_term}/i },
+               { phone_number: /#{telephone_regex(term)}/ },
                "addresses.postcode": /#{escaped_term}/i)
       }
 
+      def self.telephone_regex(term)
+        # Remove any non-digits excluding "+"
```
irisfaraway, line 100: Think this needs to be a few lines down!

irisfaraway (changes requested):
Refactor looks much cleaner, nice work! I have one tiny tiny change to suggest but then happy to approve

irisfaraway (approved):
:shipit:

PR #1263, opened 2022-09-07
Fixing the partial telephone match

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-2074

Here we have the fix for the BO telephone search mechanism. Previously it was returning partial matches, but the issue with this is if you typed in just 1 digit i.e '6' it would return any phone number that included a 6 in it resulting in a lot of matches. This partial match has been removed for telephone numbers less than 10 digits. 
Anything less than 10 digits requires a full match (meaning if a number has been entered wrongly and only contains 9 digits for example, it will still match. Anything over 10 digits will get turned into the regex string where it can be match with spaces, dashes and the +44 at the beginning as well.


irisfaraway (approved):
:shipit:

PR #1346, opened 2023-04-03
add reg_identifier to deregistration email

PaulDoyle-DEFRA:
This change adds the registration's reg_identifier to the Notify parameters in the deregistration email service.
https://eaflood.atlassian.net/browse/RUBY-2439


timstone:
@PaulDoyle-DEFRA I've created a new template with the reg_identifier in, could you use b9926a88-95db-47bd-96d4-0aaae7a322d3 instead for the id please? that way we don't have to worry about updating the existing template in the release.

timstone (dismissed):

PR #1355, opened 2023-04-25
Add code to assign an area to an address through postcode

jjromeo:


PaulDoyle-DEFRA (dismissed):

Gemfile.lock
```diff
@@ -268,7 +272,7 @@ GEM
       faraday (>= 0.9)
       sawyer (~> 0.8.0, >= 0.5.3)
     orm_adapter (0.5.0)
-    os_map_ref (0.5.0)
```
brujeo, line 271: This is bit unusual. version 0.1.1 is back from 2016. Any particular reasons for downgrading?

brujeo (approved):
Looks good to me

PR #1386, opened 2023-06-08
confirm payment method

PaulDoyle-DEFRA:
This adds a stage to the new registration and renewal journeys to confirm the chosen payment method.
https://eaflood.atlassian.net/browse/RUBY-2463

### DIFF
--- a/config/locales/forms/other_businesses_forms/en.yml
+++ b/config/locales/forms/other_businesses_forms/en.yml
@@ -1,9 +1,13 @@
 en:
   waste_carriers_engine:
-    other_businesses_forms:
+    payment_method_confirmation_forms:
       new:
-        title: Do you ever deal with waste from other businesses or households?
-        heading: Do you ever deal with waste from other businesses or households?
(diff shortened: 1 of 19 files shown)


config/locales/forms/other_businesses_forms/en.yml
```diff
@@ -1,9 +1,13 @@
...
-    other_businesses_forms:
+    payment_method_confirmation_forms:
       new:
-        title: Do you ever deal with waste from other businesses or households?
-        heading: Do you ever deal with waste from other businesses or households?
+        title: Confirm payment method
+        heading: Are you sure you want to pay by %{payment_method}?
+        paragraph: If you don't have your payment details to hand you can return to this page later. Your answers are saved for 30 days.
+        payment_method:
+          card: credit card
```
brujeo, line 9: This has always confused me. Shouldn't it say credit/debit card?  Are we taking credit cards only? Or do we assume that credit is a debit card as well?

brujeo (dismissed):

PR #1389, opened 2023-06-14
Feature/ruby 2477 wcr tech upgrade to rails 7

jjromeo:


Gemfile
```diff
@@ -16,22 +16,23 @@ gemspec
 gem "defra_ruby_template"
 
 # Use MongoDB as the database
-gem "mongoid", "~> 7.5"
+gem "mongoid"
```
PaulDoyle-DEFRA, line 19: It might be best to re-pin mongoid and devise (at least) to their new/updated major versions.

app/services/concerns/waste_carriers_engine/can_merge_finance_details.rb
```diff
@@ -12,7 +12,12 @@ def merge_finance_details
         registration.finance_details.orders << order
       end
 
+      transient_payments_array = []
```
PaulDoyle-DEFRA, line 15: Why was it necessary to break this into two steps?

app/services/waste_carriers_engine/notify/registration_confirmation_email_service.rb
```diff
@@ -33,7 +33,7 @@ def registered_address
       end
 
       def date_registered
-        @registration.metaData.date_registered.in_time_zone("London").to_date.to_s
+        @registration.metaData.date_registered.in_time_zone("London").to_date.to_date.to_s(:standard)
```
PaulDoyle-DEFRA, line 36: `.to_date` is repeated

app/services/waste_carriers_engine/renewal_completion_service.rb
```diff
@@ -97,7 +97,7 @@ def extend_expiry_date
     end
 
     def create_order_item_logs
-      OrderItemLog.create_from_registration(transient_registration)
+      OrderItemLog.create_from_registration(registration)
```
PaulDoyle-DEFRA, line 100: Shouldn't the order_item_log be created based on the new order dat in the transient_registration, not what was previously stored on the registration?

spec/support/shared_contexts/sample_registrations.rb
```diff
@@ -19,6 +18,6 @@
                    business_type: business_type,
                    tier: tier,
                    key_people: key_people,
-                   metaData: { route: route })
+                   metaData: { route: nil })
```
PaulDoyle-DEFRA, line 21: Why was it necessary to change this to nil? All registrations should have a route.

waste_carriers_engine.gemspec
```diff
@@ -16,37 +16,38 @@ Gem::Specification.new do |s|
...
 
   s.add_dependency "mongo_session_store"
   # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
   s.add_dependency "turbolinks"
   # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
-  s.add_dependency "jbuilder", "~> 2.11"
+  s.add_dependency "jbuilder"
 
   # Use AASM to manage states and transitions
-  s.add_dependency "aasm", "~> 5.3"
```
PaulDoyle-DEFRA, line 36: Same comment as for Gemfile - should some of these be pinned to their new major version number?

spec/support/shared_contexts/sample_registrations.rb
jjromeo: This is because [this test](https://github.com/DEFRA/waste-carriers-engine/blob/760fc9dbad1448ca1396f174e75048801282d9d3/spec/presenters/waste_carriers_engine/certificate_presenter_spec.rb#L135) is trying to test the metadata route being set, but that context is setting it in an implicit way that stops the test working

app/services/concerns/waste_carriers_engine/can_merge_finance_details.rb
jjromeo: The code used to be 
```
transient_registration.finance_details.payments.each do |payment|
   registration.finance_details.payments << payment
end
```

This is because the each loop was actually not working properly when called on `transient_registration.finance_details.payments`. The loop actually does not complete and I believe it's due to `transient_registration.finance_details.payments` being edited by the line inside the loop (payment can only belong to one payments collection at a time). The breaking up was a way to ensure each payment actually gets iterated on. I will add a comment to that effect

PaulDoyle-DEFRA: But the line inside the loop is acting on the registration, not on the transient_registration?

jjromeo: Yes the line is acting on the registration, but if you move a payment from `transient_registration.finance_details.payments` into `registration.finance_details.payments ` it has edited the `transient_registration.finance_details.payments` array during the iteration

app/services/waste_carriers_engine/renewal_completion_service.rb
jjromeo: So the data can be different? I added a spec to check if it would be the same

PaulDoyle-DEFRA: Yes - the registration should have all order items since it was created, so for example the items from the original registration order; some copy cards ordered a few months later; and a renewal order. The transient registration should have only the order items from the current session. I'm fairly sure that copying from the registration (if it has previous order items) would result in duplication of the older ones.

app/services/concerns/waste_carriers_engine/can_merge_finance_details.rb
jjromeo: Based on the spec that I was trying to fix, it was not copying it, but moving it instead

PaulDoyle-DEFRA (dismissed):

PR #1404, opened 2023-07-10
Feature/ruby 2533 wcr changes to certificates for numbering copies

brujeo:
Adding copy number (aka version number) to certificate
https://eaflood.atlassian.net/browse/RUBY-2533


jjromeo (dismissed):

app/models/waste_carriers_engine/registration.rb
```diff
@@ -110,8 +110,28 @@ def original_activation_date
...
+
     private
 
+    def update_certificate_version_history(version, user = nil)
+      certificate_version_history = metaData.certificateVersionHistory << {
+        version: version,
+        generated_at: DateTime.now,
+        generated_by: user&.email
+      }
+      metaData.update(certificateVersionHistory: certificate_version_history)
```
PaulDoyle-DEFRA, line 132: is the `update` necessary? Does the `<<` not persist the additional data?

app/presenters/waste_carriers_engine/certificate_presenter.rb
```diff
@@ -61,6 +61,12 @@ def certificate_creation_date
       Date.today.strftime("%e %B %Y")
     end
 
+    def certificate_version
+      return 1 if metaData.certificate_version < 1
```
PaulDoyle-DEFRA, line 65: The model defaults the version to zero. Should it be zero here, or 1 as default in the model?

brujeo: That's correct, the logic behind this was as follows:
- default to 0 in the model, because this is where counting starts. When at least one cert version is generated we increment number to 1. 
 - certificate_version method is used to display cert version number. And that is always have to be greater than 0. Defaulting to 1 is added as an exception handling when we've reset cert number to 0, but haven't incremented it. This should never happen with current ACs, but it was possible when I've only started to work on this task  - e.g. when certificate version gets reset to 0 during renewal, but then cert gets generated from back office. We now increment version numbers from both front and back offices,

PaulDoyle-DEFRA: Why not return zero if no certificate has been generated yet? Returning one implies that a certificate has been generated when it has not been.

PaulDoyle-DEFRA (approved):

jjromeo (approved):

app/presenters/waste_carriers_engine/certificate_presenter.rb
brujeo: I agree that this might sound bit confusing. However, this method is being used inside the certificate template. I thought that showing text like "This is copy number 0 of the certificate." would be inappropriate. And as I've mentioned above we had an unlikely, but possible use case when the cert version resets to 0, and then cert gets generated from back office. This use case is not valid anymore as CBD decided that cert versions should be incremented when generating certs both from front and back offices.

PaulDoyle-DEFRA (approved):

PR #1453, opened 2023-10-20
Refactor to remove duplicate upper_tier definition

PaulDoyle-DEFRA:
https://eaflood.atlassian.net/browse/RUBY-2091


app/controllers/waste_carriers_engine/govpay_forms_controller.rb
```diff
@@ -14,7 +14,7 @@ def new
         flash[:error] = I18n.t(".waste_carriers_engine.govpay_forms.new.setup_error")
         go_back
       else
-        redirect_to payment_info[:url]
+        redirect_to payment_info[:url], allow_other_host: true
```
jjromeo, line 17: Is this change not potentially significant?

PaulDoyle-DEFRA: Without this change one of the unit tests fails - not clear to me why this has only started to fail now:
rspec ./spec/requests/waste_carriers_engine/govpay_forms_spec.rb:62
`Unhandled exception: Unsafe redirect to "https://www.payments.service.gov.uk/secure/c3b8908b-68df-478a-ad94-6a724e614e46", pass allow_other_host: true to redirect anyway.`

jjromeo (approved):

PR #1457, opened 2023-10-26
Remove mailcatcher / smtp / action_mailer code and config

PaulDoyle-DEFRA:
https://eaflood.atlassian.net/browse/RUBY-2756


spec/dummy/config/environments/test.rb
```diff
@@ -26,11 +26,6 @@
   # Disable request forgery protection in test environment.
   config.action_controller.allow_forgery_protection = false
 
-  # Tell Action Mailer not to deliver emails to the real world.
-  # The :test delivery method accumulates sent emails in the
-  # ActionMailer::Base.deliveries array.
-  config.action_mailer.delivery_method = :test
```
brujeo, line 32: This is questionable. We also don't want emails to be sent via notify either when running tests. What will happen to emails when that option gets removed?

brujeo (approved):

PR #1481, opened 2023-12-13
Chore/ruby 2770 move bo functionality to bo

PaulDoyle-DEFRA:
This change moves several back-office-specific classes out of the engine, to be moved to the back-office application.
https://eaflood.atlassian.net/browse/RUBY-2801


app/controllers/concerns/waste_carriers_engine/can_redirect_form_to_correct_path.rb
```diff
@@ -13,6 +13,9 @@ def redirect_to_correct_form
       # new_state_name_path/:token or start_state_name_path?token=:token
       def form_path
         send("new_#{@transient_registration.workflow_state}_path".to_sym, token: @transient_registration.token)
+      rescue NoMethodError
+        # if unable to find the route in the engine, let the app try to resolve it
+        main_app.send("new_#{@transient_registration.workflow_state}_path".to_sym, token: @transient_registration.token)
       end
```
brujeo, line 19: Is this expected behaviour or is this rescue added just as a safety measure? 

- If we're expecting this rescue to be hit in normal circumstances, probably better to refactor it with if-else

- If we're not expecting the rescue to be hit in normal circumstances, then some error logging here would be handy.

spec/support/shared_examples/request_get_flexible_form.rb
```diff
@@ -7,71 +7,64 @@
 # If the user loads one of these forms, we change the workflow_state to match their request,
 # so the browser and the database agree about what form the user is currently using.
 RSpec.shared_examples "GET flexible form" do |form|
-  context "when a valid user is signed in" do
-    let(:user) { create(:user) }
 
-    before do
-      sign_in(user)
+  context "when a renewal is in progress" do
+    let(:transient_registration) do
```
brujeo, line 12: I see you have removed `sign_in(user)` from many tests. Is it because this is purely front-office functionality where users no longer need to sign-in?

PaulDoyle-DEFRA: Yes. Ultimately we should be able to remove the devise and cancancan gems from the engine. It's not quite there yet.

brujeo (dismissed):

brujeo (approved):

PR #1497, opened 2024-01-29
Revert 'Outside England' / 'Not found' change

PaulDoyle-DEFRA:
https://eaflood.atlassian.net/browse/RUBY-2819

### DIFF
--- a/spec/services/waste_carriers_engine/determine_area_service_spec.rb
+++ b/spec/services/waste_carriers_engine/determine_area_service_spec.rb
@@ -37,8 +37,8 @@ module WasteCarriersEngine
         context "with no match found" do
           let(:response) { instance_double(DefraRuby::Area::Response, successful?: false, error: DefraRuby::Area::NoMatchError.new) }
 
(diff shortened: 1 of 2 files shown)


jjromeo (approved):

spec/services/waste_carriers_engine/determine_area_service_spec.rb
```diff
@@ -37,8 +37,8 @@ module WasteCarriersEngine
         context "with no match found" do
           let(:response) { instance_double(DefraRuby::Area::Response, successful?: false, error: DefraRuby::Area::NoMatchError.new) }
 
-          it "returns 'Not found'" do
-            expect(described_class.run(coordinates)).to eq "Not found"
+          it "returns 'Outside England'" do
+            expect(described_class.run(coordinates)).to eq "Outside England"
```
brujeo, line 41: Not sure if that's correct to say that address is Outside England if NoMAtchError is returned. I believe same error is returned when address lookup just fails

brujeo (approved):

PR #1506, opened 2024-02-21
Feature/ruby 2846 certificate versioning

PaulDoyle-DEFRA:
* Change default certificate version from 0 to 1
* Increment certificate version only when created/renewed, not when viewed
https://eaflood.atlassian.net/browse/RUBY-2846

### DIFF
--- a/app/models/waste_carriers_engine/registration.rb
+++ b/app/models/waste_carriers_engine/registration.rb
@@ -110,25 +110,21 @@ module WasteCarriersEngine
     end
 
     def increment_certificate_version(user = nil)
-      new_version = metaData.certificateVersion.to_i.succ
-      metaData.update(certificateVersion: new_version)
+      # If history is empty, this is the first version so keep the default value
+      new_version = metaData.certificateVersionHistory.empty? ? 1 : metaData.certificateVersion.to_i.succ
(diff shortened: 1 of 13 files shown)


app/models/waste_carriers_engine/registration.rb
```diff
@@ -110,25 +110,21 @@ def original_activation_date
     end
 
     def increment_certificate_version(user = nil)
-      new_version = metaData.certificateVersion.to_i.succ
-      metaData.update(certificateVersion: new_version)
+      # If history is empty, this is the first version so keep the default value
+      new_version = metaData.certificateVersionHistory.empty? ? 1 : metaData.certificateVersion.to_i.succ
```
jjromeo, line 114: Shouldn't certificateVersion be 1 by default ? Why does it hardcode 1 instead of taking the current certificateVersion value?

jjromeo (approved):

app/models/waste_carriers_engine/registration.rb
PaulDoyle-DEFRA: Yes, it should be 1 (the default value) if version history is empty. The code just worked a little more cleanly this way, skipping an update of 1 with 1.

PR #1558, opened 2024-08-07
Fix/ruby 3145 missing refund update logic

PaulDoyle-DEFRA:
This change moves two govpay services from the back-office to the engine and uses them to process refund webhook calls.
https://eaflood.atlassian.net/browse/RUBY-3145

### DIFF
--- a/app/services/waste_carriers_engine/govpay_update_refund_status_service.rb
+++ b/app/services/waste_carriers_engine/govpay_update_refund_status_service.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class GovpayUpdateRefundStatusService < WasteCarriersEngine::BaseService
+
+    def run(registration:, refund_id:, new_status:)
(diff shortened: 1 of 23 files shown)


app/services/waste_carriers_engine/govpay_update_refund_status_service.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module WasteCarriersEngine
+  class GovpayUpdateRefundStatusService < WasteCarriersEngine::BaseService
+
+    def run(registration:, refund_id:, new_status:)
+      return false if new_status != "success"
```
brujeo, line 7: minor: it might be a good idea to store these statuses as registration class constants (or payment/refund class constants if we have such dedicated classes) 

e.g. 

`new_status != WasteCarriersEngine::Payment::PAYMENT_STATUS_SUCCESS`

brujeo (approved):

brujeo (approved):

PR #1574, opened 2024-10-09
Chore/ruby 3409 mongoid reference one

PaulDoyle-EA:
This removes an obsolete custom `can_reference_single_document_in_collection` association workaround in favour of more conventional Rails/mongoid structures.
It also retires the use of `company_address` on the Address model, in favour of the existing and preferred `registered_address`. Note that no data changes are required as part of this change.
https://eaflood.atlassian.net/browse/RUBY-3409


app/models/concerns/waste_carriers_engine/can_have_registration_attributes.rb
```diff
@@ -23,11 +22,30 @@ module CanHaveRegistrationAttributes
...
+          unless address.blank? || address.address_type == address_type.upcase
+            raise ArgumentError, "Attempted to set #{address_type} address with address of type #{address.address_type}"
+          end
+
+          # clear any existing address of this type
+          addresses.where(address_type: address_type.upcase).first&.destroy!
+
+          addresses << address
+        end
+      end
```
brujeo, line 48: I don't think this matches what was there before.
Previously, we had 

contact_address -> address_type: POSTAL
company_address -> address_type: REGISTERED
registered_address -> address_type: REGISTERED

Now we have

contact_address -> address_type: POSTAL
postal_address -> address_type: POSTAL
registered_address -> address_type: REGISTERED

Please double check if code is correct

PaulDoyle-EA: I agree postal_address is redundant. The rest looks correct - company_address is no more.

brujeo (approved):

PR #1609, opened 2025-02-25
Fix/ruby 3459 payment callbacks

PaulDoyle-EA:
This change fixes two defects:
1. Payments were not being created in the system when a valid govpay payment callback indicated that the payment had a pending status.
2. Payment balance was not being updated when a valid govpay webhook callback updated payment status to "success".
https://eaflood.atlassian.net/browse/RUBY-3459

### DIFF
--- a/spec/services/waste_carriers_engine/govpay_webhook_refund_service_spec.rb
+++ b/spec/services/waste_carriers_engine/govpay_webhook_refund_service_spec.rb
@@ -30,6 +30,9 @@ module WasteCarriersEngine
 
       let(:update_refund_service) { instance_double(WasteCarriersEngine::GovpayUpdateRefundStatusService) }
 
+      # # Make finance details refundable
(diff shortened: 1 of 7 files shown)


spec/services/waste_carriers_engine/govpay_webhook_refund_service_spec.rb
```diff
@@ -87,6 +90,19 @@ module WasteCarriersEngine
...
+              # context "when the webhook changes the status to a non-success value" do
+
+              context "when the webhook changes the status to success" do
+                let(:prior_payment_status) { Payment::STATUS_SUBMITTED }
+
+                before { assign_webhook_status("success") }
+
+                it "updates the balance" do
+                  expect { run_service }.to change { wcr_payment.finance_details.reload.balance }
+                end
```
brujeo, line 104: this check gets repeated across multiple tests, consider moving it to shared_examples

brujeo (approved):

PR #1628, opened 2025-06-03
Update payment callback and next_url logic

PaulDoyle-EA:
https://eaflood.atlassian.net/browse/RUBY-3754

### DIFF
--- a/app/services/waste_carriers_engine/govpay_payment_service.rb
+++ b/app/services/waste_carriers_engine/govpay_payment_service.rb
@@ -13,7 +13,6 @@ module WasteCarriersEngine
 
     def prepare_for_payment
       Rails.logger.tagged("GovpayPaymentService", "prepare_for_payment") do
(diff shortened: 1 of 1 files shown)


app/helpers/waste_carriers_engine/defra_ruby_govpay_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+        mocks_url_var = "WCRS_MOCK_BO_GOVPAY_URL"
+        external_url_var = "WCRS_PUBLIC_DOMAIN"
+      else
+        mocks_url_var = "WCRS_MOCK_FO_GOVPAY_URL"
+        external_url_var = "WCRS_BACK_OFFICE_DOMAIN"
+      end
+
+      mocks_url = ENV.fetch(mocks_url_var)
+      mocks_url_root = url_root(mocks_url)
+      external_url = ENV.fetch(external_url_var)
```
brujeo, line 33: ENV.fetch() will raise an error if variable not set. Shall we set some defaults here in case it's not set? Alternatively, defaults can be set in the initializer for defra_ruby_mocks

PR #1718, opened 2026-06-11
Fix/ruby 4317 wcr govpay investigate callback errors

brujeo:
Fixed 2 GovPay Handler errors:
- Mongoid::Errors::DocumentNotFound
- invalid govpay_id

### DIFF
--- a/app/services/waste_carriers_engine/govpay_payment_webhook_handler.rb
+++ b/app/services/waste_carriers_engine/govpay_payment_webhook_handler.rb
@@ -6,8 +6,9 @@ module WasteCarriersEngine
 
     def run(webhook_body)
       @webhook_body = webhook_body
-      @payment_id = webhook_body.dig("resource", "payment_id")
-      @payment = GovpayFindPaymentService.run(payment_id: payment_id)
+      @payment = find_payment
+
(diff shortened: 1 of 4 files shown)


app/services/waste_carriers_engine/govpay_payment_webhook_handler.rb
```diff
@@ -30,6 +31,26 @@ def run(webhook_body)
       raise
     end
 
+    # A payment document is only persisted once a payment succeeds, so a
+    # missing payment is only a concern if the webhook reports success.
```
jjromeo, line 35: With the change making missing payments only raise when the status is success, are we ok with no longer raising in cases where the status is nil, a status we do not recognise, or the payload is incorrectly formed.

brujeo: There's a different check that checks whether the payload is correct. This change is only about not being able to find a payment for a valid reason. 

Currently, we do not store unsuccessful/cancelled/errored payments. So, when we receive a webhook callback request for such a payment, we can't find the payment object. This was causing an error. Now, it will only raise error if can't find the payment object for a successful payment.

jjromeo (approved):

PR #1735, opened 2026-07-24
[RUBY-4350] Add geospatial EA area lookup using MongoDB

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4350

Contains the model and services for repeating strategy implemented in WEX and FRAE, so area lookups can query against the actual boundaries, replacing the bounding-box matching used by the old external area look-up. In this case adding an EaPublicFaceArea model storing EA public face area boundaries as GeoJSON (WGS84).

There was a conversion of coordinates needed from easting / northing to WGS84 due to mongodb only supporting the latter when it comes to queries, this is done using the breasal gem. DetermineEaAreaService validates coordinates and returns the containing area name.

In this case no Ops change needed (unlike WEX's PostGIS install)

The data load (load_ea_areas rake task) and the swap from the API lookup will be in back-office PR.


waste_carriers_engine.gemspec
```diff
@@ -73,6 +73,10 @@ Gem::Specification.new do |s|
   # Used to build and parse XML requests
   s.add_dependency "nokogiri", "~> 1.19.1"
 
+  # Used to convert British National Grid eastings and northings to WGS84
+  # latitude and longitude for MongoDB geospatial queries
+  s.add_dependency "breasal", "~> 0.0.1"
```
brujeo: I don't think it's a good idea to rely on a archived gem that's still at version 0.0.1 and has been updated for 13 years. Instead, we could add a conversion service to one of our repositories (e.g. defra-ruby-address), which would also make it available for use by our other services.

brujeo (approved):

